PR #1, opened 2022-11-21
docs(react-boundary): yarn install should be add keyword

ChanhyukPark-Tech:
Thank you for the good library.
There was a small bug, so I fixed it😁


manudeli (approved):
Great, @ChanhyukPark-Tech Thanks for first PR!

PR #3, opened 2022-12-31
fix(react): remove unnecessary type in fallback

### LINKED ISSUE #2: BaseErrorbounday fallback type is something wrong in @suspensive/react
### Relevant code or config

```ts
fallback: ReactNode | ((props: { error: Error; reset: (...args: unknown[]) => void }) => ReactNode)
```
### Problem description:
fallback type has reset prop and it take arg as a factor. but It's not used anywhere

### Suggested solution: 
i think it works too
```ts
fallback: ReactNode | ((props: { error: Error; reset: () => void }) => ReactNode)
```

jinn2u:
### remove unnecessary type in fallback
I changed the factor of fallback's reset function as void 

fix #2

### DIFF
--- a/packages/react/src/ErrorBoundary.tsx
+++ b/packages/react/src/ErrorBoundary.tsx
@@ -22,7 +22,7 @@ type Props = PropsWithRef<
     resetKeys?: unknown[]
     onReset?(): void
     onError?(error: Error, info: ErrorInfo): void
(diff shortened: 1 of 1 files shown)


manudeli:
@jinn2u, Thanks!

manudeli:
@all-contributors please add @jinn2u for bug

PR #19, opened 2023-02-12
ci: check typescript, lint, test code on PR synchronized

ChanhyukPark-Tech:
# Overview

Check the various test when PR is generated and updated. If the CI pipeline works for each commit, it will take a lot of time to refine the commit, so let the pipeline work for each PR😄

Originally, I used this [library](https://github.com/toss/yarn-plugin-workspace-since) to detect only modified package itself to execute the command of the various check, so that only the package where the change occurred would be various check.

However, I did not use the above [library](https://github.com/toss/yarn-plugin-workspace-since) because I thought that the `suspensive` uses [turborepo](https://turbo.build/) and the package that did not change due to the nature of [turborepo](https://turbo.build/) would omit the check.

If the various check works unnecessarily on a package that has not even changed, I will improve it by using [library](https://github.com/toss/yarn-plugin-workspace-since)

### DIFF
--- a/.github/workflows/typescript-workflow.yml
+++ b/.github/workflows/typescript-workflow.yml
@@ -10,6 +10,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Install pnpm
+        run: npm install -g pnpm
+
       - name: Install dependencies
         run: pnpm install
 
--- a/.github/workflows/typescript-workflow.yml
+++ b/.github/workflows/typescript-workflow.yml
@@ -1,0 +1,26 @@
+name: 'Type checking for typescript'
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  test:
+    name: 'Validating Type'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/cache@v2
+        id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+        with:
+          path: .yarn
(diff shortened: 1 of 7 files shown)


.github/workflows/typescript-workflow.yml
```diff
@@ -0,0 +1,20 @@
...
+    name: 'Validating Type'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install pnpm
+        run: npm install -g pnpm
+
+      - name: Install dependencies
+        run: pnpm install
```
manudeli, line 17: I want to remove repetitively installing pnpm, dependencies in https://github.com/suspensive/react/pull/20/files#diff-e894bfa63f52fe487b10fce0d90f8276523f2a08f5120025aaafe3e5c72a5d23R13-R17, How about merge this file with eslint-workflow.yml in https://github.com/suspensive/react/pull/20 and rename this file like `ci-workflow.yml`? and add workflow's job for testing code by jest after?

I think it would be better if we test all(type:check, test by jest, eslint) in ci-workflow.yml at once, with allowing only one time to install pnpm, dependencies in ci-workflow. How do you think this?

ChanhyukPark-Tech: I thought it would be better to divide it to make sure that we know where we failed.
After read your comment, I think it's better to run it in the same workflow for performance!
Also, if we go into the [action page](https://github.com/suspensive/react/actions), We will see where our code failed in detail, so it would be good to merge with the same file! I'll fix it😃

manudeli (approved):

manudeli:
@ChanhyukPark-Tech, Thanks!👍

PR #23, opened 2023-02-16
test(react-testing-library): add initial test configuration for react

ChanhyukPark-Tech:
# Overview

thank you for your good suggestion!

I didn't have enough time, so I proceeded with the minimum setting. It was confirmed that the dom test was also successful with the `pnpm test` command.


configs/jest/jest.config.js
```diff
@@ -134,7 +134,7 @@ module.exports = {
   // setupFiles: [],
 
   // A list of paths to modules that run some code to configure or set up the testing framework before each test
-  // setupFilesAfterEnv: [],
+  setupFilesAfterEnv: ['<rootDir>/jest.setup.js'],
```
ChanhyukPark-Tech, line 137: It refers jest.setup.js in your root folder

packages/react-query/jest.setup.js
```diff
@@ -0,0 +1 @@
+import '@testing-library/jest-dom'
```
ChanhyukPark-Tech, line 1: It makes we can use react-dom test in our test file

manudeli (approved):
Thanks! 👍

PR #27, opened 2023-03-09
test(react): write test code for hooks (`useIsMounted`, `useKey`)

ChanhyukPark-Tech:
# 🎁Congratulation this pr is @manudeli birthday present🥳

# Overview

write test code for useIsMounted, useKey

1. useIsMounted : check value is true when mounted
2. useKey : check key is updated when execute refresh function

# Review

in useKey I want code like below because it is more readability 
but result, current destructing not have updated value so can't use destructing

```tsx

describe('useKey', () => {
  describe('refresh function', () => {
    it('should update key state', async () => {
      const {
        result: {
          current: [key, refresh],   // this point key is not updated 
        },
      } = renderHook(() => useKey())

      expect(key).toBe(0)

      act(() => {
        refresh()
      })
      expect(key).toBe(1)
    })
  })
})

```

## PR Checklist


packages/react/src/hooks/useKey/useKey.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+      })
+      expect(result.current[0]).toBe(INITIAL_VALUE + 1)
+
+      act(() => {
+        result.current[1]()
+      })
+      expect(result.current[0]).toBe(INITIAL_VALUE + 2)
+    })
+  })
+})
```
manudeli, line 35: How about this? I remove unnecessary `async` keyword, and unnecessary INITIAL_VALUE

```suggestion
import { act, renderHook } from '@testing-library/react'
import useKey from './useKey'

describe('useKey', () => {
  describe('refresh function', () => {
    it('should increment the key state by 1', () => {
      const { result } = renderHook(() => useKey())
      const initialKey = result.current[0]
      const refresh = result.current[1]
      expect(initialKey).toBe(0)

      act(refresh)
      expect(result.current[0]).toBe(initialKey + 1)

      act(refresh)
      expect(result.current[0]).toBe(initialKey + 2)
    })
  })
})
```

ChanhyukPark-Tech: Your suggestion was the best.

removed unnecessary `async` keywords.
Unnecessary constant `INITIAL_VALUE` was eliminated.

As you have shown, first declare a constant through result.current. 

In the subsequent comparison expect(result.current[0]) it is also good to approach 


Readability has been greatly improved.

manudeli (approved):
I feel so good thanks to this fun birthday present. Let's run towards 100% test coverage. thank you 👍

PR #28, opened 2023-03-10
test(react): write test code for `isDifferentArray`

ChanhyukPark-Tech:
# Overview

wrote test code for util function which name is `isDifferentArray`

I realized this while writing the test code, but this function doesn't seem to compare exactly unless it's a [primitive type](https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/Object/is). So I wrote the last test of the test code.

## PR Checklist


packages/react/src/utils/isDifferentArray/isDifferentArray.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  it('should return true when two array is same object', () => {
+    const value1 = [{ test: 1 }, { test: 2 }]
+    const value2 = [{ test: 1 }, { test: 2 }]
+
+    const result = isDifferentArray(value1, value2)
+
+    expect(result).toBe(true)
+  })
+})
```
manudeli, line 48: It's a minor difference, but I want to write a test code that is a little more concise and easy to understand, so How about indicating the primitive type and the reference type in the variable name? (ex. number, array1, object1,2)

And added some edge test cases not existing before this comment. If it looks okay to you, I would appreciate it if you could commit suggestion.

```suggestion
import { isDifferentArray } from '.'

const primitive = 0
const reference1 = { test: 0 }
const reference2 = { test: 0 }

describe('isDifferentArray', () => {
  it('should return true if the two arrays have different lengths.', () => {
    const array1 = Array.from({ length: 1 }).map((_, index) => primitive + index)
    const array2 = Array.from({ length: 2 }).map((_, index) => primitive + index)

    expect(isDifferentArray(array1, array2)).toBe(true)
  })

  it('should return false if the two arrays have same lengths and same primitive element in each index of arrays.', () => {
    const array1 = Array.from({ length: 1 }).map((_, index) => primitive + index)
    const array2 = Array.from({ length: 1 }).map((_, index) => primitive + index)

    expect(isDifferentArray(array1, array2)).toBe(false)
  })

  it('should return true if the two arrays have same lengths but at least one primitive element is different in each index of arrays.', () => {
    const array1 = [primitive, primitive + 1]
    const array2 = [primitive, primitive]

    expect(isDifferentArray(array1, array2)).toBe(true)
  })

  it('should return true if the two arrays have same lengths and have all same primitive elements but order is different', () => {
    const array1 = [primitive, primitive + 1]
    const array2 = [primitive + 1, primitive]

    expect(isDifferentArray(array1, array2)).toBe(true)
  })

  it('should return true when two arrays have each references elements in index of array have different references', () => {
    const array1 = [reference1, { test: 2 }]
    const array2 = [reference1, { test: 2 }]

    expect(isDifferentArray(array1, array2)).toBe(true)
  })

  it('should return false when two arrays have each references elements in index of array have same references', () => {
    const array1 = [reference1, reference2]
    const array2 = [reference1, reference2]

    expect(isDifferentArray(array1, array2)).toBe(false)
  })
})

```

manudeli (commented):
Oh Thanks, you're right. `different` of isDifferentArray is limited different. I didn't think about this when I created this function. I made it by referring to the comparison operation of dependencyList of useEffect hook.

Isn't there a better name for this function? If there is a better name, I think it would be nice to make it an issue and work on it in the next PR.

ChanhyukPark-Tech:
>Oh Thanks, you're right. different of isDifferentArray is limited different. I didn't think about this when I created this function. I made it by referring to the comparison operation of dependencyList of useEffect hook.
Isn't there a better name for this function? If there is a better name, I think it would be nice to make it an issue and work on it in the next PR.

It would be nice to change function name, too! 
I think we should think of a good name!

packages/react/src/utils/isDifferentArray/isDifferentArray.spec.ts
ChanhyukPark-Tech: 84~97 lines of test is really cool 😁

manudeli (approved):
Cool! I made issue to rename isDifferentArray. Let's refactor this function's name to proper thing in next Pull Request👍

PR #34, opened 2023-04-15
chore(.changeset, .github/workflows): add auto generating CHANGELOG.md

### LINKED ISSUE #16: [Feature]: I want something like CHANGELOG.md
### **Package Scope**

- [ ] New package

Package name:

### **Overview**

I'm actually using suspensive well because you provided me a good library. I'm looking forward to it more and more because you're continuing to update new features.
As a library user, I wonder what features have been added depending on the version. It would be nice if we could check the update log and so on somewhere other than PR or commit!

- github release ?

manudeli:
fix #16 

# Auto generating CHANGELOG.md
1. Using changeset bot in Pull Request
We can use changeset bot in Pull Request of every contributors to add `changeset log` that will make CHANGELOG.md
image

2. When there is a new commit on the main branch, If there is `changeset log`(inside of .changeset/*), It will make new Pull Request to bump versions of all packages. like https://github.com/manudeli/monorepo-libraries-changeset-test/pull/3 containing auto updated CHANGELOG.md files of packages

3. Then, If we merge 2.'s Pull Request, Publishing to NPM workflow will starting. like https://github.com/manudeli/monorepo-libraries-changeset-test/actions/runs/4710530002

## PR Checklist

### DIFF
--- a/.changeset/beige-ways-type.md
+++ b/.changeset/beige-ways-type.md
@@ -1,0 +1,6 @@
+---
+'@suspensive/react': patch
+'@suspensive/react-query': patch
+---
+
+changelog initialization
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -1,4 +1,4 @@
-name: "Continuous Integration"
+name: ci-workflow
 
 on:
   pull_request:
@@ -6,13 +6,21 @@ on:
 
 jobs:
   quality:
-    name: "CI: Check quality"
+    name: Check quality
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
 
       - name: Install pnpm
-        run: npm install -g pnpm
+        uses: pnpm/action-setup@v2.2.4
+        with:
+          version: 8
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.17'
+          check-latest: true
 
       - name: Install dependencies
         run: pnpm install
@@ -25,3 +33,6 @@ jobs:
 
       - name: Test by jest
         run: pnpm test
+
+      - name: Build package
+        run: pnpm run build
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -17,13 +17,13 @@ jobs:
           version: 8
 
       - name: Read .nvmrc
-         run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
-         id: nvm
+        run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
+        id: nvm
 
       - name: Setup Node
         uses: actions/setup-node@v3
         with:
-          node-version: "${{ steps.nvm.outputs.NVMRC }}"
+          node-version: ${{ steps.nvm.outputs.NVMRC }}
           check-latest: true
 
       - name: Install dependencies
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -16,10 +16,14 @@ jobs:
         with:
           version: 8
 
+      - name: Read .nvmrc
+         run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
+         id: nvm
+
       - name: Setup Node
         uses: actions/setup-node@v3
         with:
-          node-version: '16.17'
+          node-version: "${{ steps.nvm.outputs.NVMRC }}"
           check-latest: true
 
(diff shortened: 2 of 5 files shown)


.changeset/beige-ways-type.md
```diff
@@ -0,0 +1,6 @@
+---
+'@suspensive/react': patch
+'@suspensive/react-query': patch
+---
```
manudeli, line 4: We can check how changeset action will work. when we merge this Pull Request
changeset/action will make Pull Request containing CHANGELOG.md

.github/workflows/release-workflow.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: pnpm install
+
+      - name: Check typescript
+        run: pnpm type:check
+
+      - name: Check eslint
+        run: pnpm lint
+
+      - name: Test by jest
+        run: pnpm test
```
ChanhyukPark-Tech, line 37: Isn't it already verified in `ci-workflow.yml` before pushing to main branch

**I don't think a double check is necessary** because this is not a part where human error can occur, is there a reason to verify it when you push it to the main one more time?

I think it's a waste of resources from the github action runner😊

.github/workflows/ci-workflow.yml
```diff
@@ -1,18 +1,26 @@
...
-        run: npm install -g pnpm
+        uses: pnpm/action-setup@v2.2.4
+        with:
+          version: 8
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.17'
+          check-latest: true
```
ChanhyukPark-Tech, line 23:
```suggestion
      - name: Read .nvmrc
         run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
         id: nvm

      - name: Setup Node
        uses: actions/setup-node@v3
        with:
          node-version: "${{ steps.nvm.outputs.NVMRC }}"
          check-latest: true
```

I think it would be better to automatically track the number written in the `.nvmrc` of this project.

If it is used as a [`magic number`](https://drehzr.tistory.com/1136) like this, I think the management point will continue to increase!

ChanhyukPark-Tech (approved):
Thank you for safisfying exactly the requirements.

I think there are so many roles that changeset plays Thank you.

ChanhyukPark-Tech (changes requested):
😁

.github/workflows/ci-workflow.yml
manudeli: Cool suggestion

.github/workflows/release-workflow.yml
manudeli: I don't think the only way to create a new commit on the main branch is through a Pull Request.

For example, even if it's not a pull request, I think we should check the quality at least once if there's a new commit in the main branch, either by force push or by other means. Are there any other good ways to satisfy this requirement?

ChanhyukPark-Tech: I didn't think of that, but I understood! Thank you.

I think there are two ways.

1. Protect the main branch with a [protect branch](https://kotlinworld.com/292) and limit direct push (We can take more stability) In this case, you don't have to do `ci` again.

2. If you are in a situation where you concerned (Push directly through main) I think we need this `test phrase`.

I think it would be good if the code that goes into main branch only be generated through Pull Request for review and history tracking in any situation, so I think number 1 would be good! How about you💬

.github/workflows/ci-workflow.yml
ChanhyukPark-Tech: @manudeli  I commented only ci-workflow but same code has in release-workflow.yml ! 😃

.github/workflows/release-workflow.yml
manudeli: I turned protect branch rule on from now. and I remove 3 steps(Check typescript, Check eslint, Test by jest) in release-workflow.yml

.github/workflows/release-workflow.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v2.2.4
+        with:
+          version: 8
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.17
```
ChanhyukPark-Tech, line 24: It remains 😂

manudeli: I updated it, thanks for your eagle eyes!

ChanhyukPark-Tech (approved):
I learned a lot. Thank you😁

Issue #36, opened 2023-04-19
[Feature]: Prepare @tanstack/react-query v5


manudeli:
### **Package Scope**

- [x] Add to an existing package
- [ ] New package

Package name:

PR #38, opened 2023-04-28
fix(react-query): fix useSuspenseQuery, useSuspenseInfiniteQuery receive various params like original @tanstack/react-query v4

manudeli:
related #36 

# Fix useSuspenseQuery, useSuspenseInfiniteQuery receive various params like original @tanstack/react-query v4
From @tanstack/react-query v5, I saw a [BREAKING CHANGE](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#supports-a-single-signature-one-object) that the form of parameters received in the future will not be received in various forms, but only in object form.
image

But @suspensvie/react-query was still only available in the format below.
`arg1: queryKey, arg2: queryFn, arg3: Omit<UseSuspenseQueryOptions<TQueryFnData, TError, TData, TQueryKey>, 'queryKey' | 'queryFn'>`

So, we need to make the queryKey of arg1 or the queryFn of arg2 possibly included in UseSuspenseQueryOptions.

## AS-IS
```tsx
const query = useSuspenseQuery(queryKey, queryFn, options)
```
## TO-BE
```tsx
// Now we can use the same function in these different ways.
const query = useSuspenseQuery(queryKey, queryFn, options)
const query = useSuspenseQuery(queryKey, { queryFn, ...options })
const query = useSuspenseQuery({ queryKey, queryFn, ...options })
```

So I created this Pull Request.

## Expected effect when merging this Pull Request
With this Pull Request, we can hope that @suspensive/react-query users will gradually migrate from @tanstack/react-query v4 to @tanstack/react-query v5.

## Add [tsd](https://github.com/SamVerschueren/tsd) to test d.ts files built
In addition, [tsd](https://github.com/SamVerschueren/tsd) was added in this Pull Request to test that this method does not return the wrong type, so we can easily type-check the built d.ts while working.

## PR Checklist


.github/workflows/ci-workflow.yml
```diff
@@ -29,6 +29,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
+      - name: Build core packages
+        run: pnpm build --filter=@suspensive/react --filter=@suspensive/react-query
+
```
manudeli, line 34: ci-workflow.yml will be responsible for only core packages (@suspensive/react, @suspensive/react-query). So it saves time on github actions.

@suspensive/docs and @suspensive/visualization are removed from this workflow because vercel can check for errors at build time.

packages/react-query/package.json
```diff
@@ -30,39 +30,44 @@
     "build": "rm -rf dist esm && tsc -p tsconfig.build.json --declaration --emitDeclarationOnly --declarationDir dist && rollup -c rollup.config.js",
     "lint": "eslint .",
     "npm:publish": "pnpm publish",
-    "prepack": "pnpm build",
+    "prepack": "pnpm build && pnpm test:tsd",
```
manudeli, line 33: Always check the type of the d.ts file built with tsd before `pnpm pack` command to increase the stability of the package.

packages/react-query/src/useSuspenseQuery.ts
```diff
@@ -21,12 +28,13 @@ export type UseSuspenseQueryOptions<
   TError = unknown,
   TData = TQueryFnData,
   TQueryKey extends QueryKey = QueryKey
-> = Omit<UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>, 'suspense' | 'queryKey' | 'queryFn'>
+> = Omit<UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>, 'suspense'>
```
manudeli, line 31: From now on, UseSuspenseQueryOptions will also support queryKey and queryFn.

packages/react-query/src/useSuspenseInfiniteQuery.ts
```diff
@@ -30,15 +31,13 @@ export type UseSuspenseInfiniteQueryOptions<
   TError = unknown,
   TData = TQueryFnData,
   TQueryKey extends QueryKey = QueryKey
-> = Omit<
-  UseInfiniteQueryOptions<TQueryFnData, TError, TData, TQueryFnData, TQueryKey>,
-  'suspense' | 'queryKey' | 'queryFn'
->
+> = Omit<UseInfiniteQueryOptions<TQueryFnData, TError, TData, TQueryFnData, TQueryKey>, 'suspense'>
```
manudeli, line 34: From now on, UseSuspenseInfiniteQueryOptions will also support queryKey and queryFn.

manudeli (commented):
For better understanding, please leave more details in the comments.

packages/react-query/src/useSuspenseQuery.test-d.ts
```diff
@@ -0,0 +1,71 @@
+import { expectError, expectType } from 'tsd'
+import { useSuspenseQuery } from '../dist'
+
+const queryKey = ['key'] as const
+const queryFn = async () => 'response' as const
+const boolean = Math.random() > 0.5
```
ChanhyukPark-Tech, line 6: I think it would be good to declare it like this, so how about fix the `useInfiniteQuery` part separately as boolean?

packages/react-query/src/useSuspenseQuery.test-d.ts
```diff
@@ -0,0 +1,71 @@
...
+expectType<undefined>(
+  useSuspenseQuery({
+    queryKey,
+    queryFn,
+    enabled: false,
+  }).data
+)
+
+// no arg
+expectError(useSuspenseQuery())
```
ChanhyukPark-Tech, line 71: this detail is awesome..

ChanhyukPark-Tech (commented):
Thank you for responding quickly to the update!

Is there a reason why you kept the typescript version at 4?
Is 5 still unstable?😁

ChanhyukPark-Tech (approved):
If you finish this pr please merge yourself!
I approved 👍

PR #52, opened 2023-05-28
fix(react): add next.js appRouter 'use client' directive

### LINKED ISSUE #51: [Feature]: Check if we need 'use client'
### **Package Scope**
all

Package name: all

https://github.com/TanStack/query/pull/4738

manudeli:
fix #51

# Overview

I use rollup banner https://rollupjs.org/plugin-development/#banner
like https://github.com/dvtng/react-loading-skeleton/pull/162 's config


manudeli:
@ChanhyukPark-Tech 

It's clear that there are no issues with this Pull Request.
The next thing to do depends on this Pull Request. I will bypass the branch rule with administrator privileges and merge first. 

but, please review the code even after merging in order to prevent any mistakes and to align understanding.

PR #55, opened 2023-05-28
feat(react): add useSuspenseCache, suspenseCache(experimental)

### LINKED ISSUE #53: [Feature]: add useSuspenseCache, suspenseCache
### **Package Scope**

Package name: react

manudeli:
close #53 

# Add useSuspenseCache, suspenseCache(experimental) in @suspensive/react
I want to talk about this new API(useSuspenseCache, suspenseCache) with you @ChanhyukPark-Tech 

I thought that in most cases it should be easily implemented using suspense with just @suspensive/react without @tanstack/react-query. In this case, I thought there should be a new feature that makes it easy to handle functions that return promises with suspense.

## The features I'm thinking of are:
1. useSuspenseCache looking at the same cache.key should utilize only one promise and be optimized even if it is requested at the same time in the mounted component that is used.
2. useSuspenseCache looking at the same cache.key must be re-rendered together when reset is performed in the mounted component that is used. Implemented following the Observer pattern.
3. When ErrorBoundary of @suspensive/react is reset, if all error objects in suspenseCache are initialized and components that use suspenseCache are mounted, all of them must be re-rendered to cause a suspense.

I've implemented these features in advance, but I still felt there was a lot missing.

### SSR test
In order to easily test this, we migrated from the pages directory method to the AppRouter method in the previous Pull Request to see if it is well supported in the SSR environment.

I prepared it for testing at http://localhost:3001/useSuspenseCache.

ps. Thanks to @tooooo1 making me think about useSuspenseCache, suspenseCache


packages/react/src/useSuspenseCache.tsx
```diff
@@ -0,0 +1,142 @@
...
+        }),
+      onNotifies: [],
+    }
+
+    this.cache.set(key, newCache)
+    // eslint-disable-next-line @typescript-eslint/no-throw-literal
+    throw newCache.promise
+  }
+
+  public getData = <TKey extends string>(key: TKey) => this.cache.get(key)?.data
```
ChanhyukPark-Tech, line 84: what is this function purpose?🤔
it has use case?

packages/react/src/useSuspenseCache.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { useEffect, useMemo } from 'react'
+import { useRerender } from './hooks'
+
+type Tuple<T = unknown> = T[] | readonly T[]
+type Cache<Key extends string = string> = {
+  promise?: Promise<unknown>
+  key: Key
+  error?: unknown
+  data?: unknown
+  onNotifies: ((...args: unknown[]) => unknown)[]
```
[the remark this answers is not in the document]
manudeli: Oh Thanks no use case! I removed it in [b845cfa](https://github.com/suspensive/react/pull/55/commits/b845cfa92c863cff31c851a3db8069615403ceea)

packages/react/src/useSuspenseCache.tsx
ChanhyukPark-Tech: Oh it is useful😁
hot about this
```suggestion
  public getData = <TKey extends string[]>(key: TKey) => this.cache.get(JSON.stringify(key))?.data
```

because when real use I think it is so hard to use

```tsx
const customCache = useSuspenseCache({
    key: ['custom'],
    fn: () => api.manual({ successPercentage: 50, waitMs: 500 }),
  } as const)

  suspenseCache.getData(JSON.stringify(['custom']))

```

manudeli: So cool!

manudeli: @ChanhyukPark-Tech 
and I edit it as `TKey extends Tuple` not string[] in [28ac54d](https://github.com/suspensive/react/pull/55/commits/28ac54dde7ce65267e699f62d378be3aa49c687c)
I understood your good intention!

ChanhyukPark-Tech (approved):
goodgood👍

PR #60, opened 2023-05-29
fix(react-query): add @suspensive/react as peerDependenciesMeta

### LINKED ISSUE #46: [Feature]: add @suspensive/react as @suspensive/react-query's peerDependenciesMeta
### **Package Scope**

Package name: @suspensive/react-query

manudeli:
fix #46 

# Add @suspensive/react as peerDependenciesMeta of @suspensive/react-query
Users who only want to use @suspensive/react-query had to install @suspensive/react because it is a peerDependency.

I expect that allowing @suspensive/react-query to be used without installing @suspensive/react will allow more flexibility in accepting users, and as a result, @suspensive/react will be able to receive more users as well.

### peerDependenciesMeta
That's why I try to solve this problem with [peerDependenciesMeta](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependenciesmeta) in package.json.
Users of @suspensive/react-query are no longer required to install @suspensive/react.

It's nice to be flexible, but users of QueryErrorBoundary and QueryAsyncBoundary can get lost. Therefore, the official documentation and ts-doc guide users that they need to install @suspensive/react.

## PR Checklist

### DIFF
--- a/packages/react-query/src/QueryAsyncBoundary.tsx
+++ b/packages/react-query/src/QueryAsyncBoundary.tsx
@@ -44,6 +44,19 @@ if (process.env.NODE_ENV !== 'production') {
 /**
  * This component wrapping QueryErrorResetBoundary of @tanstack/react-query with @suspensive/react's AsyncBoundary.
  *
+ * So you must install @suspensive/react first, then use it.
+ * ```shell
+ * npm install @suspensive/react @suspensive/react-query
+ * ```
+ *
+ * ```shell
+ * pnpm add @suspensive/react @suspensive/react-query
+ * ```
+ *
+ * ```shell
+ * yarn add @suspensive/react @suspensive/react-query
+ * ```
+ *
  * with this component, You don't have to make unnecessary repetitive implementation to combine AsyncBoundary with QueryErrorResetBoundary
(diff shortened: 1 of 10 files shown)


packages/react-query/src/QueryAsyncBoundary.tsx
```diff
@@ -44,6 +44,19 @@ if (process.env.NODE_ENV !== 'production') {
...
+ * ```
+ *
+ * ```shell
+ * pnpm add @suspensive/react @suspensive/react-query
+ * ```
+ *
+ * ```shell
+ * yarn add @suspensive/react @suspensive/react-query
+ * ```
+ *
```
ChanhyukPark-Tech, line 59: Users watching this `ts-doc` already have `@suspensive/react-query` installed, so wouldn't it be enough to show only the `@suspensive/react` installation phrase?

```suggestion
 * So you must install @suspensive/react first, then use it.
 * ```shell
 * npm install @suspensive/react
 * ```
 *
 * ```shell
 * pnpm add @suspensive/react
 * ```
 *
 * ```shell
 * yarn add @suspensive/react
 * ```
 *
```

manudeli: 👍👍👍

ChanhyukPark-Tech (approved):
cool😄

PR #67, opened 2023-06-04
test(useRerender): add test code

manudeli:
# Overview
We want to focus on adding test code before adding other features. Make code that shouldn't change into hard code.
Because of this, when adding more suspensive features, we can immediately notice unnecessary side effects for other features.

## PR Checklist


packages/react/src/hooks/useRerender/useRerender.ts
```diff
@@ -2,8 +2,10 @@ import { useCallback, useReducer } from 'react'
...
 type Rerender = () => void
 const reducer = (state: number) => state + 1
-export const useRerender = (): Rerender => {
+const useRerender = (): Rerender => {
   const [, dispatch] = useReducer(reducer, 0)
 
   return useCallback(() => dispatch(), [])
 }
+
+export default useRerender
```
ChanhyukPark-Tech, line 11: I wonder why you used default export!
Any advantage?!😃

manudeli: I just unified with other hook.

Do you prefer named exports? (Actually I am)

Issue #62, opened 2023-05-31
[Feature]: test all correctly with coverage


manudeli:
### **Package Scope**

- [x] Add to an existing package

Package name:

ChanhyukPark-Tech:
I will make test code for [Delay](https://github.com/suspensive/react/blob/main/packages/react/src/Delay.tsx) Component

manudeli:
I'll close this issue because we achieved 100% test coverage🎉
Thanks to you guys @okinawaa @tooooo1

PR #75, opened 2023-06-08
test(react): add test case for Suspense

manudeli:
related #62

# Add test case for Suspense of @suspensive/react

I added test case.

## PR Checklist

### DIFF
--- a/packages/react/src/Suspense.spec.tsx
+++ b/packages/react/src/Suspense.spec.tsx
@@ -49,6 +49,16 @@ describe('Suspense', () => {
     jest.advanceTimersByTime(ms)
     await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument())
   })
+  it('.CSROnly should render the fallback during suspense', () => {
+    render(
+      <Suspense.CSROnly fallback={FALLBACK}>
+        {TEXT}
+        <ThrowDuring ms={ms} />
+      </Suspense.CSROnly>
+    )
+    expect(screen.queryByText(FALLBACK)).toBeInTheDocument()
+    expect(screen.queryByText(TEXT)).not.toBeInTheDocument()
+  })
   it('.CSROnly should render the children after the suspense', async () => {
     jest.useFakeTimers()
     render(
@@ -57,10 +67,10 @@ describe('Suspense', () => {
         <ThrowDuring ms={ms} />
       </Suspense.CSROnly>
     )
-    expect(screen.queryByText(FALLBACK)).toBeInTheDocument()
     needThrow = false
     jest.advanceTimersByTime(ms)
     await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument())
+    expect(screen.queryByText(FALLBACK)).not.toBeInTheDocument()
   })
   it('.CSROnly should render the children if nothing to suspend in children', async () => {
     render(<Suspense.CSROnly fallback={FALLBACK}>{TEXT}</Suspense.CSROnly>)
--- a/packages/react/src/Suspense.spec.tsx
+++ b/packages/react/src/Suspense.spec.tsx
@@ -1,0 +1,70 @@
+import { render, screen, waitFor } from '@testing-library/react'
+import { Suspense } from './Suspense'
+
+const ms = 1000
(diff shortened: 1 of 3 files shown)


packages/react/src/Suspense.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+  })
+  it('.CSROnly should render the children after the suspense', async () => {
+    jest.useFakeTimers()
+    render(
+      <Suspense.CSROnly fallback={FALLBACK}>
+        {TEXT}
+        <ThrowDuring ms={ms} />
+      </Suspense.CSROnly>
+    )
+    expect(screen.queryByText(FALLBACK)).toBeInTheDocument()
```
chanhyukpark-29cm, line 60: Is it right that this phrase should be verified in this `it`section?

this section is about `.CSROnly should render the children after the suspense` 
I think this phrase should be seperated other `it` section 😃

### example

```tsx
  it('.CSROnly should render the fallback during suspense', async () => {
    expect(screen.queryByText(FALLBACK)).toBeInTheDocument()


```

chanhyukpark-29cm: I suggested one test in one `it` section because I thought it would be good to verify only one feature😄

packages/react/src/Suspense.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+    needThrow = false
+    jest.advanceTimersByTime(ms)
+    await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument())
+  })
+  it('.CSROnly should render the children if nothing to suspend in children', async () => {
+    render(<Suspense.CSROnly fallback={FALLBACK}>{TEXT}</Suspense.CSROnly>)
+    expect(screen.queryByText(FALLBACK)).not.toBeInTheDocument()
+    expect(screen.queryByText(TEXT)).toBeInTheDocument()
+  })
```
chanhyukpark-29cm, line 69: cool!!

packages/react/src/Suspense.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+const Throw = () => {
+  throw new Promise((resolve) => resolve('resolved'))
+}
+
+describe('Suspense', () => {
+  let needThrow = true
+  beforeEach(() => {
+    needThrow = true
+  })
+  const ThrowDuring = (props: { ms: number }) => {
```
chanhyukpark-29cm, line 17: I think it may cause misunderstanding that the word `props` is used for react components, so why don't you modify it as follows?

```suggestion
  const ThrowDuring = (params: { ms: number }) => {
```

chanhyukpark-29cm (commented):
thx for great job 👍

ChanhyukPark-Tech:
I posted a review without knowing it was my company account😁
It'll be okay, right?

packages/react/src/Suspense.spec.tsx
manudeli: image

ThrowDuring should be used as a react component. If ThrowDuring is just used as a function, an error like the screenshot above may occur. So I think the name props is correct.

If you want to check it yourself, you can see what happens if you fix it like params and run ThrowDuring as a function instead of ThrowDuring react component. I don't think ThrowDuring should be used as a function because Suspense can catch the child component's pending promise.

Therefore, I think it should be used like ReactComponent with props rather than params.

packages/react/src/Suspense.spec.tsx
manudeli: Testing one by one separately. cool!

manudeli:
@ChanhyukPark-Tech Not a big deal in my opinion, but it would be nice if it could be avoided. kk
I reflected all of your code reviews. Let me know if you have any other ideas [88fe14f](https://github.com/suspensive/react/pull/75/commits/88fe14f05865ce91804a259cadc27e915aadac8b)

ChanhyukPark-Tech (approved):

packages/react/src/Suspense.spec.tsx
ChanhyukPark-Tech: oh I totally understood by reading your cool comment 👍

PR #79, opened 2023-06-12
fix(react): fix useErrorBoundaryGroup to guarantee parent ErrorBoundaryGroup

manudeli:
# Fix useErrorBoundaryGroup to guarantee parent ErrorBoundaryGroup

When we use useErrorBoundaryGroup, we could use it even though there is no ErrorBoundaryGroup as parent component. However, in the future it will throw an error in this case.

When this Pull Request is merged, I hope to avoid forgetting to use ErrorBoundaryGroup during development.

### I added test case more
- ErrorBoundaryGroup.spec.tsx
- ErrorBoundary.spec.tsx

## PR Checklist

### DIFF
--- a/.changeset/thick-walls-talk.md
+++ b/.changeset/thick-walls-talk.md
@@ -1,0 +1,5 @@
+---
+"@suspensive/react": patch
+---
+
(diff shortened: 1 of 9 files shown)


ChanhyukPark-Tech (approved):
I'm getting better at writing test codes
It feels good to be documented.

PR #83, opened 2023-06-16
fix(react): rename useSuspenseCache to useAwait, suspenseCache to suspensiveCache

### LINKED ISSUE #86: [Feature] AwaitClient's methods need to work with index of key with hierarchy
> currently, we use stringified key yet, but next Pull Request, I want to change it suspenseCache's methods need to work with index of key with hierarchy like queryKey of @tanstack/query

I think this is need to be implemented in this Pull Request. so I change this Pull Request as draft. after that, I'll request review

_Originally posted by @manudeli in https://github.com/suspensive/react/issues/83#issuecomment-1596098279_

manudeli:
related #62 

# Rename useSuspenseCache to Await, suspenseCache to suspensiveCache
- useSuspenseCache -> useAwait
- suspenseCache -> suspensiveCache

ps. I added test code for Await, useAwait

## PR Checklist


manudeli:
> currently, we use stringified key yet, but next Pull Request, I want to change it suspenseCache's methods need to work with index of key with hierarchy like queryKey of @tanstack/query

I think this is need to be implemented in this Pull Request. so I change this Pull Request as draft. after that, I'll request review

packages/react/src/experimental/index.ts
```diff
@@ -0,0 +1 @@
+export { Await, useAwait } from './Await'
```
manudeli, line 1: We can use experimental features like this

```tsx
import { Await, useAwait } @suspensive/react/experimental
```

manudeli (commented):
I commented my intentions

websites/visualization/src/app/react/experimental/await/page.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <ErrorBoundary fallback={() => <>error</>}>
+      <Suspense fallback={<>loading...</>}>
+        <Await options={{ key: [2000] as const, fn: asyncFn }}>
+          {(awaited) => (
+            <div>
+              <div>{awaited.data}</div>
+              <div>{awaited.data}</div>
+            </div>
+          )}
+        </Await>
```
manudeli, line 26: The reason Cache was removed from the name is that `useQuery` of @tanstack/react-query also uses Cache internally, but uses the more important name `Query` to express querying not `Cache`. Therefore, SuspenseCache was also renamed to focus on `awaiting promise` with the corresponding fn `Await`, not `Cache`.

packages/react/src/__test__/experimental/Await.spec.tsx
```diff
@@ -0,0 +1,111 @@
+import { act, render, screen, waitFor } from '@testing-library/react'
```
okinawaa, line 1: It's not related to this PR, but I have a question!

Is it better to be in a folder with only a test code?
Isn't it better to be as close to the actual implementation as possible according to the principle of locality when maintaining? To look at the implementation and the test code together

manudeli: I understood you may mean removing __test__ directory is better
I'll remove __test__ and refactor it after this PR merging or close

packages/react/src/experimental/suspensiveCache.tsx
```diff
@@ -0,0 +1,113 @@
...
+          newCache.error = error
+        }),
+    }
+
+    this.cache.set(stringifiedKey, newCache)
+    // eslint-disable-next-line @typescript-eslint/no-throw-literal
+    throw newCache.promise
+  }
+
+  public getData = <TKey extends Tuple>(key: TKey) => this.cache.get(JSON.stringify(key))?.data
```
okinawaa, line 79: Don't we need getPromise or getError?!😄

packages/react/src/experimental/suspensiveCache.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  private keyNotifiesMap = new Map<string, ((...args: unknown[]) => unknown)[]>()
+
+  public attach<TKey extends Tuple>(key: TKey, onNotify: (...args: unknown[]) => unknown) {
+    const stringifiedKey = JSON.stringify(key)
+    const keyNotifies = this.keyNotifiesMap.get(stringifiedKey)
+    this.keyNotifiesMap.set(stringifiedKey, [...(keyNotifies ?? []), onNotify])
+
+    const detachSelf = () => this.detach(key, onNotify)
+    return detachSelf
```
okinawaa, line 89:
```suggestion
    return { detach } 
```

I think it takes some learning to think the `detachSelf` function with the return of attach. It was a bit unpredictable!(only my think!..😄)
 
Why don't we return it in the form of an object so that the user can see the function name and figure out the role?

packages/react/src/experimental/Await.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+type AwaitProps<TData, TKey extends Tuple> = {
+  options: AwaitOptions<TData, TKey>
+  children: FunctionComponent<Awaited<TData>>
+}
+
+/**
+ * @experimental This is experimental feature.
+ */
+export const Await = <TData, TKey extends Tuple>({ children: Children, options }: AwaitProps<TData, TKey>) => {
```
okinawaa, line 45: The function of `detach` is defined in the `suspensiveCache`, but I wonder if it is not necessary to use it in the Await component

packages/react/src/experimental/suspensiveCache.tsx
```diff
@@ -0,0 +1,113 @@
...
+  private notifyToAttacher = <TKey extends string>(key?: TKey) =>
+    key
+      ? this.keyNotifiesMap.get(key)?.forEach((notify) => notify())
+      : this.keyNotifiesMap.forEach((keyNotifies) => keyNotifies.forEach((notify) => notify()))
+}
+
+/**
+ * @experimental This is experimental feature.
+ */
+export const suspensiveCache = new SuspensiveCacheObserver()
```
okinawaa, line 113: cool 👍

websites/visualization/src/app/CommonLayout.tsx
```diff
@@ -10,7 +10,7 @@ export const CommonLayout = ({ children }: PropsWithChildren) => {
   return (
     <>
       <TopNavigation />
-      <Link href="/react/useSuspenseCache">🔗 Experimental Feature: useSuspenseCache</Link>
+      <Link href="/react/await">🔗 Experimental Feature: Await</Link>
```
okinawaa, line 13:
```suggestion
      <Link href="/react/experimental/await">🔗 Experimental Feature: Await</Link>
```

websites/visualization/src/app/react/experimental/await/page.tsx
okinawaa, line 26: Will children be invisible before `asyncFn` is resolved? 
Does the external `Suspense` handle the loading(pending) process?

packages/react/src/experimental/suspensiveCache.tsx
```diff
@@ -0,0 +1,113 @@
+import { Tuple } from '../types'
+
+type Cache<Key extends Tuple = Tuple> = {
+  promise?: Promise<unknown>
+  key: Key
+  error?: unknown
+  data?: unknown
+}
+
+class SuspensiveCacheObserver {
```
okinawaa, line 10: Doesn't the forced reload(cmd + R) of Chrome blow the cache?
Even if I force a reload, the cached data remains.

packages/react/src/experimental/Await.tsx
manudeli: `detach` will be invoked on `useAwait`'s unmount internally

README.md
```diff
@@ -38,7 +38,9 @@ All declarative components to use suspense on both CSR, SSR.
 - Delay
 - SuspensiveProvider
 - HOC(Higher Order Component)s
-- useSuspenseCache, suspenseCache (Experimental)
+- useAwait, Await (Experimental)
+
+- HOC(Higher Order Component)s
```
manudeli, line 43:
```suggestion
- useAwait, Await (Experimental)
```

websites/visualization/src/app/react/experimental/await/page.tsx
manudeli: Yeah, That's right! and this will be cached in cachekey. we can reuse

packages/react/src/experimental/suspensiveCache.tsx
manudeli: Let's make another features after this PR

packages/react/src/experimental/suspensiveCache.tsx
manudeli: chrome will blow cache on reload. did you test it in Next.js 13? if right, I guess you meet already suspensed data in server without meeting Suspense's fallback

packages/react/src/experimental/suspensiveCache.tsx
manudeli: Beautiful! I think it will make `suspensive.attach` more extendable by object with naming clarity

packages/react/src/experimental/suspensiveCache.tsx
okinawaa: Oh i understood
It works in server i think!

okinawaa (approved):
Cool

.changeset/plenty-mayflies-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@suspensive/react': patch
```
manudeli, line 2:
```suggestion
'@suspensive/react': minor
```

PR #85, opened 2023-06-16
fix: add eslint rule to detect dead code

tooooo1:
# Overview

Introducing ESLint conventions to the Suspensive library by discovering some deprecated code in the Suspensive code base.

## Changes made
- Added rule `no-unused-vars` to ESLint configuration
- Adjusted existing code to follow this rule

## How does this PR impact users?
- This PR does not impact users directly but improves the code quality and maintainability by preventing unused code from being committed to the project.

## AS-IS
image

## TO-BE
image

## PR Checklist

### DIFF
--- a/configs/eslint-config-ts/noimport.js
+++ b/configs/eslint-config-ts/noimport.js
@@ -5,5 +5,6 @@ module.exports = {
   plugins: ['@typescript-eslint'],
   rules: {
     '@typescript-eslint/no-empty-function': ['off'],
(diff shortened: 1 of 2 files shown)


manudeli (approved):
Thanks👍 @ChanhyukPark-Tech How about this rule to you

ChanhyukPark-Tech (approved):
I think it's a very good improvement. If import it even though we don't use it, it can cause misunderstanding when other devloper read the code

PR #92, opened 2023-07-03
fix(react): remove ErrorBoundaryGroup's not on purpose situation group.reset because of isMounted's side effect 

manudeli:
# Remove ErrorBoundaryGroup's not on purpose situation group.reset because of isMounted's side effect 

### AS-IS
ErrorBoundaryGroup will trigger reset on render because of useIsMounted

### TO-BE
ErrorBoundaryGroup won't trigger reset on render because of removed useIsMounted
And to keep feature of ErrorBoundary's reset by parent ErrorBoundaryGroup, I added usePrevious to compare previous parent ErrorBoundaryGroup's resetKey with current parent ErrorBoundaryGroup's resetKey

## PR Checklist


manudeli:
@minsoo-web I checked this Pull Request you made https://github.com/toss/slash/pull/277

I want to show you this Pull Request to remove this useIsMounted of @toss/react in @toss/error-boundary
When I made this ErrorBoundaryGroup in @toss/errorBoundary, I used useIsMounted but I think it is unnecessary hook now.

Check this please and could you review this Pull Request

packages/react/src/hooks/useIsMountedOnThisRender/useIsMountedOnThisRender.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  useEffect(() => {
+    isMountedOnThisRenderRef.current = true
+
+    return () => {
+      isMountedOnThisRenderRef.current = false
+    }
+  }, [])
+
+  return isMountedOnThisRenderRef.current
```
minsoo-web, line 14: I think this value is not changed for using this hook situation.

Because, it`s return value is never change until re-call

for example, I guess like this

```js
let FOO = 1;
const foo = () => FOO;
const bar = () => {
  FOO = 2;
}

const result = foo(); // 1;
bar();
// result is also 1;

const result2 = foo();
// result2 is 2;
```

So, i think it is better return value is not current value, but refObject

```suggestion
  return isMountedOnThisRenderRef
```

plz check my test code result

minsoo-web: image

image

packages/react/src/hooks/useIsMountedState/useIsMountedState.ts
```diff
@@ -1,6 +1,6 @@
 import { useEffect, useState } from 'react'
 
-const useIsMounted = () => {
+export const useIsMountedState = () => {
```
minsoo-web, line 3: In my opinion, this hooks name is more readable for useIsMounted (original, or useCheckMounted)
Why you add sufix `State` to this hook's name?

packages/react/src/hooks/useIsMountedOnThisRender/useIsMountedOnThisRender.ts
minsoo-web: As the role of this hook provides has been mounted at this point(calling this hook), why don't you return a function that returns whether it has been mounted?

like this

```ts
export function useIsMounted() {
  const isMountedRef = useRef(true);
  const isMounted = useCallback(() => isMountedRef.current, []);

  useEffect(() => {
    return () => void (isMountedRef.current = false);
  }, []);

  return isMounted;
}
```

[ref](https://stackoverflow.com/questions/58979309/checking-if-a-component-is-unmounted-using-react-hooks)

minsoo-web (commented):
Thank you for giving chance reviewing this PR
It is my glad to review 🙇

packages/react/src/hooks/useIsMountedState/useIsMountedState.ts
manudeli: This comment is not valid now because of new commit b1d1a5c removing mounting check hook by usePrevious.

but I want to leave comment clarifing what my intention of previous code was

The reason for adding the subfix State to the name of the useIsMountedState hook is

1. useIsMountedOnThisRender: This hook only checks the mounted status of the component with ref and does not trigger rerender.
2. useIsMountedState: This hook changes the mounted state of the component and even rerenders it.

I wanted to clearly distinguish the roles of these two hooks by naming. An important distinction criterion is rerender.

In @suspensive/react we have to check the mounted state of the component in two ways.
1. Supsense.CSROnly should "rerender" and show children only when the component is mounted, but
2. ErrorBoundaryGroup only needs to check whether it is mounted or not, without rerender.

The reason I made this Pull Request was that the name useIsMounted was not enough to express the functionality of these two required hooks.

In two popular libraries, [@toss/react](https://slash.page/libraries/react/react/src/hooks/useIsMounted.i18n) and [usehooks-ts](https://usehooks-ts.com/react-hook/use-is-mounted), useIsMounted has two different functions as mentioned above.

As result, I thought there was no reason for a hook named useIsMounted to have a state and fire a rerender. Therefore, I would like to divide the names into useCheckMounted same with usehooks-ts's useIsMounted and useIsMountedState same with @toss/react's useIsMounted.

packages/react/src/hooks/useIsMountedOnThisRender/useIsMountedOnThisRender.ts
manudeli: ErrorBoundaryGroup don't need mount checking hook after b1d1a5c. but Thanks for your comment to make me care this hook carefully. I perfectly understood what you want me to how handle react's ref.

manudeli:
@minsoo-web, after your comment, I thought about whether I need really this useIsMounted or even useIsMounted with ref.

So I added [b1d1a5c](https://github.com/suspensive/react/pull/92/commits/b1d1a5c3c8d7993cbb17c6351c06801f391ae2ae) removing useIsMounted in ErrorBoundaryGroup, and using new usePrevious

I think this way is more clear to use parent group's resetKey in ErrorBoundaryGroup than before. I'm curious about your opinion and @ChanhyukPark-Tech too. could you review again please?

minsoo-web:
> I think this way is more clear to use parent group's resetKey in ErrorBoundaryGroup than before. I'm curious about your opinion and @ChanhyukPark-Tech too. could you review again please?

It is my pleasure, I will check it 🏃.

packages/react/src/hooks/usePrevious/usePrevious.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const usePrevious = <T>(value: T): T => {
+  const ref = useRef<T>(value)
+
+  useEffect(() => {
+    ref.current = value
+  }, [value])
+
+  return ref.current
+}
```
minsoo-web, line 11: It is so COOL!

My previous review that return value will not change for refObj.current until hook is re-call is applied for this hook by so convenient and readable.

It is my code [sand-box link](https://codesandbox.io/s/misty-bash-hccjcm?file=/src/App.tsx) to understand usePrevious's purpose perfectly. 

It is my suggestion for you.

What about make hook for combining useState with usePrevious ?
My imagine API's interface is below

```tsx
const [state, setState, prevState] = useNewState(0)
```

The reason I suggested this is because I don't think `usePrevious` hook will be used often in itself, but if you want to check the previous value, assuming that the update is done with `useEffect` or `useState`, it will be used often!

minsoo-web (approved):
Thank you for applying my review 🙇‍♂️
This review allowed me to think a lot about re-renders. 

It was such an honor.

packages/react/src/ErrorBoundaryGroup.tsx
```diff
@@ -28,14 +28,14 @@ export const ErrorBoundaryGroup = ({
...
 
-  const isMounted = useIsMounted()
-  const group = useContext(ErrorBoundaryGroupContext)
   useEffect(() => {
-    if (isMounted && !blockOutside) {
+    if (!blockOutside && prevParentGroupResetKey !== parentGroup?.resetKey) {
       reset()
     }
-  }, [group?.resetKey, isMounted, reset])
+  }, [prevParentGroupResetKey, parentGroup?.resetKey, reset, blockOutside])
```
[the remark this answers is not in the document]
manudeli: Cool 👍

packages/react/src/hooks/usePrevious/usePrevious.ts
manudeli: > const [state, setState, prevState] = useNewState(0)

After seeing your comment, I tried a lot to create an interface for that hook, but I think it was very difficult for me to name the hook. I also worry about whether Hook is playing too many roles. anyway I understood why you suggest it too.

If this hook is created, it is doubtful that it will be used in more cases than usePrevious. I wonder if there are any cases you are thinking of.

## New Suggestion: useIsChanged

After I saw your suggestion, I thought of a hook that can be used in more cases than `usePrevious`. I've named it `useIsChanged`. it looks like `ErrorBoundaryGroup`could use this hook as well.

```tsx
const useIsChanged(value: unknown){
  const previousValue = usePrevious(value)
  
  return previousValue !== value
}
```

### useIsChanged in ErrorBoundaryGroup
Isn't it more intuitive?

```tsx
export const ErrorBoundaryGroup = ({
  blockOutside = false,
  children,
}: {
  /**
   * If you use blockOutside as true, ErrorBoundaryGroup will protect multiple ErrorBoundaries as its children from external ErrorBoundaryGroup's resetKey
   * @default false
   */
  blockOutside?: boolean
  /**
   * Use multiple ErrorBoundaries inside of children
   */
  children?: ReactNode
}) => {
  const [resetKey, reset] = useKey()
  const parentGroup = useContext(ErrorBoundaryGroupContext)
  const isChangedParentGroupResetKey = useIsChanged(parentGroup?.resetkey)

  useEffect(() => {
    if (!blockOutside && isChangedParentGroupResetKey) {
      reset()
    }
  }, [isChangedParentGroupResetKey, reset, blockOutside])

  const value = useMemo(() => ({ reset, resetKey }), [reset, resetKey])

  return <ErrorBoundaryGroupContext.Provider value={value}>{children}</ErrorBoundaryGroupContext.Provider>
}
```

minsoo-web: > After seeing your comment, I tried a lot to create an interface for that hook, but I think it was very difficult for me to name the hook.

I deeply sympathize with what you say.
My suggestion hook is not only do TOO many things but also difficult intuitive naming.

> If this hook is created, it is doubtful that it will be used in more cases than usePrevious. I wonder if there are any cases you are thinking of.

As you understand my suggestion, my suggestion is for user can combining `useState` hook with `usePrevious`
Here is my example case.

1. I want to use state.
2. I want use previous state not using setter by callback function's args
3. so I wrote two lines of code (using `useState` and `usePrevious`)

Here is my sample code

[code-sandbox](https://codesandbox.io/s/misty-bash-hccjcm?file=/src/App.tsx)

Such example is common case of like logs.
But I recognize it is not common case,,, so I understand what you're worried about.

So, my suggestion for combining `useState` with `usePrevious` hook use case is not only common, but also can useless

minsoo-web: > New Suggestion: useIsChanged

I think it is COOL, intuitive, and good for DX!!

ChanhyukPark-Tech (commented):
It seems clear to reset it compared to the previous value, not to check with mount. I think it's more declarative.

Please use the "useIsChanged" hook. I'll review it again Thank you.😄

packages/react/src/ErrorBoundaryGroup.tsx
```diff
@@ -29,13 +29,13 @@ export const ErrorBoundaryGroup = ({
 }) => {
   const [resetKey, reset] = useKey()
   const parentGroup = useContext(ErrorBoundaryGroupContext)
-  const previousParentGroup = usePrevious(parentGroup)
+  const isChangedParentGroupResetKey = useIsChanged(parentGroup?.resetKey)
```
ChanhyukPark-Tech, line 32: why not using this variable name?!

```tsx
const `isParentGroupResetKeyChanged` = useIsChanged(parentGroup?.resetKey)
```

is...Changed is more appropriate I think! than isChanged....

manudeli: Could you use suggest code of github features. I want to commit your suggestion.
image

packages/react/src/ErrorBoundaryGroup.tsx
```diff
@@ -28,14 +28,14 @@ export const ErrorBoundaryGroup = ({
...
 
-  const isMounted = useIsMounted()
-  const group = useContext(ErrorBoundaryGroupContext)
   useEffect(() => {
-    if (isMounted && !blockOutside) {
+    if (!blockOutside && isChangedParentGroupResetKey) {
       reset()
     }
-  }, [group?.resetKey, isMounted, reset])
+  }, [isChangedParentGroupResetKey, reset, blockOutside])
```
ChanhyukPark-Tech, line 38: image

I think it can't be done for deleting line  (I knew today 😆)
image

manudeli: Ahha, Okay!

manudeli: I renamed `isChangedParentGroupResetKey` to `isParentGroupResetKeyChanged` and added you as co-author in [d9cd0a8](https://github.com/suspensive/react/pull/92/commits/d9cd0a8c44f0a5456919e19d49ec6b83bbf01b25)

packages/react/src/ErrorBoundaryGroup.tsx
manudeli: I renamed `isChangedParentGroupResetKey` to `isParentGroupResetKeyChanged` and added you as co-author in [d9cd0a8](https://github.com/suspensive/react/pull/92/commits/d9cd0a8c44f0a5456919e19d49ec6b83bbf01b25)

ChanhyukPark-Tech (approved):
@manudeli cool!

packages/react/src/hooks/usePrevious/usePrevious.ts
manudeli: > So, my suggestion for combining `useState` with `usePrevious` hook use case is not only common, but also can useless

Maybe you expect usePrevious's return value will be maintained as prev even though when component continuous rerendering. but, If component is rerendered, previousValue will be just previous Rendered value.

But I think it's good idea what you want. I want to imagine how to implement more it later.

PR #101, opened 2023-07-28
chore(bundler): migrate from rollup to tsup

### LINKED ISSUE #99: [Feature]: add tsup
### **Package Scope**

- [x] New package

Package name:

okinawaa:
# Overview

fix #99 

 
We need to change in `bundler` of libraries for consistency between library projects. `tsup` is easy to use, and is suitable to our conditions. 

## PR Checklist


packages/react-query/package.json
```diff
@@ -27,7 +27,7 @@
     "esm"
   ],
   "scripts": {
-    "build": "rm -rf dist esm && tsc -p tsconfig.build.json --declaration --emitDeclarationOnly --declarationDir dist && rollup -c rollup.config.js",
+    "build": "tsup src/index.ts --format cjs,esm --dts --minify",
```
okinawaa, line 30: we have to support cjs!?

For a better `node` ecosystem, I want to support only `esm`, but I'm also worried that certain people might prefer `cjs`

manudeli: I think we need to support cjs yet. because there is creact-react-app, jest(most popular depenency for every node developer are using) that support only cjs officially, not [esm that is experimental supported yet](https://jestjs.io/docs/ecmascript-modules) yet.

for example, swiper(support only esm) have thes issues
1. https://swiperjs.com/react#usage-with-create-react-app
2. https://github.com/nolimits4web/swiper/issues/4871
3. https://github.com/nolimits4web/swiper/discussions/4969

I want to promote library users use esm too, but at the same time, I don't want to make library users go away from us by forcing them use esm

manudeli:
I made https://github.com/suspensive/react/pull/102 to improve this Pull Request. review it please

packages/react-query/package.json
okinawaa: I'm fully agreed with your opinion😄

manudeli:
I made https://github.com/suspensive/react/pull/105 to improve this Pull Request again to support multiple entry. review it please

manudeli (approved):
I changed base branch of this Pull Request. because we don't want to make unnecessary BREAKING CHANGE by using changeset's prerelease.

[They recommand to make independant branch from `main` branch](https://github.com/changesets/changesets/blob/main/docs/prereleases.md). so I did this

After proving beta version is working in production, I'll make Pull Request to merge beta to main.

PR #105, opened 2023-07-31
fix(tsup): add multiple entry

manudeli:
related #101 

# Add multiple entry

support multiple entry

## AS IS

```ts
export const options: Options = {
  banner: { js: '"use client"' },
  format: ['cjs', 'esm'],
  entry: ['src/index.ts'],
  sourcemap: true,
  clean: true,
  dts: true,
}
```

image

## TO BE

```ts
export const options: Options = {
  banner: { js: '"use client"' },
  format: ['cjs', 'esm'],
  entry: ['src/index.ts', 'src/*.ts', 'src/*.tsx'],
  sourcemap: true,
  clean: true,
  dts: true,
  splitting: false,
}
```

image

## PR Checklist


packages/react-query/package.json
```diff
@@ -26,7 +26,7 @@
   "module": "dist/index.mjs",
   "types": "dist/index.d.ts",
   "scripts": {
-    "build": "tsup src/index.ts",
+    "build": "tsup",
```
manudeli, line 29: Moved all options to @suspensive/tsup

packages/react/src/Delay.tsx
```diff
@@ -1,5 +1,3 @@
-'use client'
```
manudeli, line 1: all directive to @suspensive/tsup's banner

manudeli (commented):
I leave comments as my intention

okinawaa (approved):

packages/react-query/package.json
okinawaa: cool

PR #114, opened 2023-08-18
fix(typescript): update version from v4 to v5

### LINKED ISSUE #33: [Feature]: new Typescript version(v5) support!
### **Package Scope**

- [ ] New package

Package name: entire package, root typescript version

### **Overview**

Type script version 5 was actually released, so it would be good to apply it.

Upgrade entire typescript version

manudeli:
close #33 

# Update TypeScript version from v4 to v5 
- I updated version of TypeScript
- I remained TypeScript only in root. and I removed it in all packages

## PR Checklist

### DIFF
--- a/.changeset/green-bobcats-flash.md
+++ b/.changeset/green-bobcats-flash.md
@@ -1,0 +1,6 @@
+---
+"@suspensive/react-query": patch
+"@suspensive/react": patch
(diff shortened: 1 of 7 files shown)


okinawaa (approved):
Thank you!

PR #129, opened 2023-09-03
ci(vitest): install dependency

manudeli:
# Overview

## PR Checklist

- [ ] I have written documents and tests, if needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,8 @@
     "prettier": "^2.8.8",
     "publint": "^0.2.2",
     "turbo": "latest",
-    "typescript": "^5.1.6"
+    "typescript": "^5.1.6",
+    "vite": "^4.4.9",
+    "vitest": "^0.34.3"
   }
 }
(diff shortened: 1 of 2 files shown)


packages/react-query/vitest.config.ts
```diff
@@ -0,0 +1,13 @@
...
+export default defineConfig({
+  test: {
+    name: '@suspensive/react-query',
+    dir: './src',
+    watch: false,
+    environment: 'jsdom',
+    globals: true,
+    setupFiles: './test.setup.ts',
+    coverage: { provider: 'v8' },
+  },
```
okinawaa, line 12: we don't have to use alias?!

I think It could possible needed

```tsx
 alias: {
      '~': '/src',
      
    },

```

okinawaa (approved):
awesome

The interface between rtl and vitest is 99.99%. It's really good

packages/react-query/vitest.config.ts
manudeli: You mean [this `alias` api](https://vitest.dev/config/#alias). I didn't know it.
I will try to know it. wait a moment

PR #130, opened 2023-09-04
feat(react): remove AsyncBoundary

manudeli:
# Remove AsyncBoundary

1. AsyncBoundary is a component that just combines Suspense and ErrorBoundary, but it does not provide any additional functions.
2. Because the number of management points increases due to this AsyncBoundary, I think it is a barrier to entry for users to use the `suspensive library`. (pendingFallback = fallback of suspense, rejectedFallback = fallback of ErrorBoundary, AsyncBoundary uses ErrorBoundary internally, so users should be aware that it is affected by ErrorBoundaryGroup and etc.)
3. Rather than using AsyncBoundary, which is a combination of Suspense and ErrorBoundary, I think it is a better library to provide the user with an interface to directly combine Suspense ErrorBoundary so that the positions of ErrorBoundary and Suspense can be changed flexibly.

## PR Checklist


okinawaa:
>AsyncBoundary is a component that just combines Suspense and ErrorBoundary, but it does not provide any additional functions.

I agree with this, but, 
I think there is a reason why the purpose of using the library is to use predefined codes **comfortably**, even though it can be implemented by themselves.

Based on what you said, `AsyncBoundary` is just a simple combination, but I know that this combination is a frequently used use case, so wouldn't it be okay to keep it as it is?

okinawaa:
As for asynchronous processing, I want to provide an environment where we can develop it comfortably by using only our suspensive.

I want to save developers the trouble of processing asynchronous processing utility on their projects.(ex. making AsyncBoundary themselves)

manudeli:
> As for asynchronous processing, I want to provide an environment where we can develop it comfortably by using only our suspensive.
> 
> I want to save developers the trouble of processing asynchronous processing utility on their projects.(ex. making AsyncBoundary themselves)

After our conversation in Discord with @okinawaa, I understood frequent use cases. first I closed this Pull Request.  if need to remove AsyncBoundary really, reopen this pull request at that time

PR #131, opened 2023-09-05
feat(vitest): add @suspensive/vitest generating named config automatically by packageJson.name

manudeli:
# @suspensive/vitest
I added @suspensive/vitest generating named config automatically by packageJson.name
So we don't have to name manually. It will find package.json's name field automatically and use it as vitest's name.

## PR Checklist


packages/react/vitest.config.ts
```diff
@@ -1,13 +1,10 @@
...
-    coverage: { provider: 'v8' },
-  },
-})
+export default defineConfig(
+  forPackage({
+    test: {
+      name: 'suspensive renamed test name',
+    },
+  })
+)
```
okinawaa, line 10: good example!

```suggestion
export default defineConfig(forPackage())
```

manudeli: kkkkkkkk Thx! 👍

manudeli:
@okinawaa I will change base branch

okinawaa (approved):

PR #142, opened 2023-09-11
fix(react): Improve Await

### LINKED ISSUE #80: [BUG]: fix useSuspenseCache tearing with startTransition, adopt useSyncExternalStore
### **Package Scope**

### **Overview**

https://github.com/reactwg/react-18/discussions/86

Implement fixing with compatiable react 17 or under with this library
https://www.npmjs.com/package/use-sync-external-store

#### Reference
react-redux's solution https://github.com/reduxjs/react-redux/blob/master/package.json#L77

### LINKED ISSUE #149: [BUG]: AwaitOption key's field order is matter because of JSON.stringify care field's order
### **Package Scope**

manudeli:
fix #80
fix #149 

# Improve Await

## 1. Replace useEffect to use-sync-external-store
I replaced useEffect to resolve tearing issue on Concurrent Rendering for External Store Libraries like us [This React Conf Session by Daishi Kato](https://conf.reactjs.org/stage) inspire me why we change useEffect to sync our awaitClient to react
image

## 2. Use hashKey 
To ignore object field's order for key of AwaitOptions
So I added hashKey and test code for it

## PR Checklist


packages/react/src/experimental/Await.tsx
```diff
@@ -1,164 +1,168 @@
...
-  data?: unknown
-}
+export const Await = <TData, TKey extends Key>({ children, options }: AwaitProps<TData, TKey>) =>
+  createElement(children, useAwait<TData, TKey>(options))
 
-class SuspensiveCacheObserver {
-  private cache = new Map<string, Cache>()
+class AwaitClient {
+  private cache = new Map<ReturnType<typeof hashKey>, AwaitState>()
+  private syncsMap = new Map<ReturnType<typeof hashKey>, ((...args: unknown[]) => unknown)[]>()
```
tooooo1, line 62:
```suggestion
  private syncsMap = new Map<ReturnType<typeof hashKey>, Sync[]>()
```

packages/react/src/utils/hashKey.ts
```diff
@@ -0,0 +1,44 @@
...
+          .reduce((result, key) => {
+            result[key] = val[key]
+            return result
+          }, {} as any)
+      : val
+  )
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+const isPlainObject = (o: any): o is Object => {
```
tooooo1, line 17: about **hashing**, wouldn't it be more rigorous to check with `type PlainObject = Record<string, any>` instead of just an object type?

```suggestion
type PlainObject = Record<string, any>

export const hashKey = (key: Key) => {
  return JSON.stringify(key, (_, val) =>
    isPlainObject(val)
      ? Object.keys(val)
          .sort()
          .reduce((result: PlainObject, key) => {
            result[key] = val[key]
            return result
          }, {})
      : val
  )
}

const isPlainObject = (o: any): o is PlainObject => {
```

we can also eliminate the lint directive `// eslint-disable-next-line @typescript-eslint/ban-types`.

packages/react/src/utils/hashKey.ts
```diff
@@ -0,0 +1,44 @@
...
+  // eslint-disable-next-line no-prototype-builtins
+  if (!prot.hasOwnProperty('isPrototypeOf')) {
+    return false
+  }
+
+  // Most likely a plain Object
+  return true
+}
+
+const hasObjectPrototype = (o: any) => Object.prototype.toString.call(o) === '[object Object]'
```
tooooo1, line 44:
```suggestion
const hasObjectPrototype = (o: PlainObject) => Object.prototype.toString.call(o) === '[object Object]'
```

packages/react/src/utils/hashKey.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  // If has modified prototype
+  const prot = ctor.prototype
+  if (!hasObjectPrototype(prot)) {
+    return false
+  }
+
+  // If constructor does not have an Object-specific method
+  // eslint-disable-next-line no-prototype-builtins
+  if (!prot.hasOwnProperty('isPrototypeOf')) {
```
tooooo1, line 36:
```suggestion
  if (!Object.prototype.hasOwnProperty.call(prot, 'isPrototypeOf')) {
```

By using Object.prototype.hasOwnProperty.call(prot, 'isPrototypeOf'), you've adhered to best practices, ensuring that the method call is safe even if hasOwnProperty is not directly present on the object. As a result, the `// eslint-disable-next-line no-prototype-builtins` comment became redundant and was correctly removed.

packages/react/src/experimental/Await.tsx
manudeli: Oh I missed it

packages/react/src/utils/hashKey.ts
manudeli: So cool!

packages/react/src/utils/hashKey.ts
manudeli: I think hasObjectPrototype don't have to constraint parameter as only `PlainObject` so I didn't follow this suggestion but Good Advice! Thanks!

tooooo1 (approved):

okinawaa (changes requested):
Cool

okinawaa (approved):

PR #151, opened 2023-09-16
fix(react): fix throw null

### LINKED ISSUE #141: [BUG]: throw null as error case
### **Package Scope**

manudeli:
fix #141 
related with #62

# ErrorBoundary should catch null thrown by children
I also added test case for it

## PR Checklist


packages/react/src/utils/toTest.tsx
```diff
@@ -29,6 +29,19 @@ export const ThrowError = ({ message, after, children }: ThrowErrorProps) => {
...
+
+type ThrowNullProps = PropsWithChildren<{ after: number }>
+export const ThrowNull = ({ after, children }: ThrowNullProps) => {
+  const [isNeedError, setIsNeedError] = useState(throwErrorIsNeed.current)
+  if (isNeedError) {
+    throw null
+  }
+  useEffect(() => {
+    setTimeout(() => setIsNeedError(true), after)
+  }, [after])
```
tooooo1, line 41:
```suggestion
  useEffect(() => {
    const timerId = setTimeout(() => setIsNeedError(true), after)

    return () => {
      clearTimeout(timerId)
    }
  }, [after])
```

Cleanup 😁

manudeli: Cool!

tooooo1 (approved):

okinawaa (approved):

PR #162, opened 2023-09-22
test(react): modify assertion test code.

okinawaa:
# Overview

I left comment to describe what I changed

## PR Checklist

### DIFF
--- a/packages/react/src/utils/assert.spec.ts
+++ b/packages/react/src/utils/assert.spec.ts
@@ -20,6 +20,7 @@ function get(type: 'foo' | 'bar') {
 }
 
 describe('assert', () => {
+  // eslint-disable-next-line vitest/expect-expect
   it('should assert condition', () => {
     const data = get('foo')
     assert(data.foo1 === 'foo1', 'dummy-message')
--- a/packages/react/src/utils/assert.spec.ts
+++ b/packages/react/src/utils/assert.spec.ts
@@ -1,16 +1,17 @@
+/* eslint-disable vitest/expect-expect */
 import { assert } from '.'
 
-function get(type: 'Helper' | 'Core') {
+function get(type: 'foo' | 'bar') {
   switch (type) {
-    case 'Helper':
+    case 'foo':
       return {
-        type: 'Helper',
-        library: 'Suspensive',
+        foo1: 'foo1',
+        foo2: 'foo2',
       } as const
-    case 'Core':
+    case 'bar':
       return {
-        type: 'Core',
-        library: 'React',
+        bar1: 'bar1',
+        bar2: 'bar2',
       } as const
     default: {
       throw new Error('no type')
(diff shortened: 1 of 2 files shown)


packages/react/src/utils/assert.spec.ts
```diff
@@ -1,16 +1,17 @@
...
+        foo1: 'foo1',
+        foo2: 'foo2',
       } as const
-    case 'Core':
+    case 'bar':
       return {
-        type: 'Core',
-        library: 'React',
+        bar1: 'bar1',
+        bar2: 'bar2',
```
okinawaa, line 14: It's better that the test code doesn't contain the domain context as much as possible.(for dummy object)

This is because it can cause confusion to people who read test code. Originally, there was a problem that the person reading the test code read the unnecessary domain contents. I replaced it with foo and bar so that I can understand it without thinking.

packages/react/src/utils/assert.spec.ts
```diff
@@ -20,13 +21,14 @@ function get(type: 'Helper' | 'Core') {
...
+    assert(data.foo1 === 'foo1', 'dummy-message')
+    expectTypeOf(data.foo2).toMatchTypeOf<string>()
+    expectTypeOf(data.bar1).not.toMatchTypeOf<string>()
   })
 
   it('should throw error if given condition is not met', () => {
-    const value = 'Suspensive' as string
-    expect(() => assert(value === 'React', "value should be 'React'")).toThrow("value should be 'React'")
+    const value = 'baz' as string
+    expect(() => assert(value === 'paz', 'condition is false')).toThrow('condition is false')
```
manudeli, line 32: How about this

```suggestion
    expect(() => assert(value === 'paz', "value should be 'paz'")).toThrow("value should be 'paz'")
```

packages/react/src/utils/assert.spec.ts
```diff
@@ -1,16 +1,17 @@
+/* eslint-disable vitest/expect-expect */
```
okinawaa, line 1: I added it because there is no expect phrase in the first test phrase.

packages/react/src/utils/assert.spec.ts
```diff
@@ -20,13 +21,14 @@ function get(type: 'Helper' | 'Core') {
 
 describe('assert', () => {
   it('should assert condition', () => {
-    const data = get('Helper')
-    assert(data.type === 'Helper', "data.type should be 'Helper'")
-    expect(data.library).toBe('Suspensive')
+    const data = get('foo')
+    assert(data.foo1 === 'foo1', 'dummy-message')
```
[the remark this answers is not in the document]
manudeli: This is better informing what is assert's spec than before👍

packages/react/src/utils/assert.spec.ts
```diff
@@ -20,13 +21,14 @@ function get(type: 'Helper' | 'Core') {
 
 describe('assert', () => {
   it('should assert condition', () => {
```
manudeli, line 23: I think this is better to scope down it.

```suggestion
describe('assert', () => {
  // eslint-disable-next-line vitest/expect-expect
  it('should assert condition', () => {
```

packages/react/src/utils/assert.spec.ts
manudeli, line 1:
```suggestion
```

packages/react/src/utils/assert.spec.ts
manudeli: So much better!! Thanks to make our library avoid confusing to contributors

manudeli (approved):

PR #164, opened 2023-09-22
refactor(react): apply assert

manudeli:
# Overview

I applied assert in use case

I removed "vitest/globals" in tsconfig to prevent contributors use `assert of chai used by vitest's global.d.ts` instead of our `assert in src/utils` So I want to force contributor write where assert import from.

### What situation I want to prevent
No error when using `assert of chai used by vitest's global.d.ts` in ErrorBoundaryGroup.tsx, not *.spec.tsx

image

## PR Checklist

### DIFF
--- a/packages/react-query/tsconfig.json
+++ b/packages/react-query/tsconfig.json
@@ -3,6 +3,6 @@
   "include": ["."],
   "exclude": ["**/*.test-d.ts", "**/*.test-d.tsx"],
   "compilerOptions": {
-    "types": ["vitest/globals", "@testing-library/jest-dom"]
+    "types": ["@testing-library/jest-dom"]
   }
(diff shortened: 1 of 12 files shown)


packages/react-query/tsconfig.json
```diff
@@ -3,6 +3,6 @@
   "include": ["."],
   "exclude": ["**/*.test-d.ts", "**/*.test-d.tsx"],
   "compilerOptions": {
-    "types": ["vitest/globals", "@testing-library/jest-dom"]
+    "types": ["node", "@testing-library/jest-dom"]
```
manudeli, line 6: "node" to use process.env.*

okinawaa (approved):

PR #166, opened 2023-09-23
test(react): add tests for withErrorBoundary

tooooo1:
related with #62

# Overview

Introducing tests for the `withErrorBoundary` ensures that is working as expected and provides the necessary error boundaries for the components it wraps. The `withErrorBoundary` is designed to catch rendering errors in its child component and display a **fallback UI** instead of crashing the entire app.

AS-IS
스크린샷 2023-09-23 오후 10 05 20

TO-BE
스크린샷 2023-09-23 오후 10 05 32

## PR Checklist

### DIFF
--- a/packages/react/src/ErrorBoundary.spec.tsx
+++ b/packages/react/src/ErrorBoundary.spec.tsx
@@ -167,7 +167,7 @@ describe('withErrorBoundary', () => {
     ThrowError.reset()
   })
 
-  it('should render the wrapped component when there`s no error', () => {
+  it("should render the wrapped component when there's no error", () => {
     const WrappedComponent = withErrorBoundary(() => <>{TEXT}</>, {
       fallback: (caught) => <>{caught.error.message}</>,
     })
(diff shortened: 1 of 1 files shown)


packages/react/src/ErrorBoundary.spec.tsx
```diff
@@ -159,3 +159,40 @@ describe('ErrorBoundary', () => {
...
+  })
+
+  it('should render the wrapped component when there`s no error', () => {
+    const WrappedComponent = withErrorBoundary(() => <>{TEXT}</>, {
+      fallback: (caught) => <>{caught.error.message}</>,
+    })
+
+    act(() => root.render(<WrappedComponent />))
+    expect(container.textContent).toBe(TEXT)
+  })
```
manudeli, line 177:
```suggestion
  it("should render the wrapped component when there's no error", () => {
    const WrappedComponent = withErrorBoundary(() => <>{TEXT}</>, {
      fallback: (caught) => <>{caught.error.message}</>,
    })

    act(() => root.render(<WrappedComponent />))
    expect(container.textContent).toBe(TEXT)
  })
```

manudeli (approved):

manudeli:
Thanks, We almost code coverage 100%!

PR #169, opened 2023-09-24
test(react): add Await, AwaitClient's clearError case

tooooo1:
# Overview

- Updated the tests for the `Await` component to verify its behavior with the `useAwait` hook.
- Added tests for the `AwaitClient` class's `clearError` method to check its behavior after a failed suspend operation.

### AS-IS
스크린샷 2023-09-24 오후 7 41 47

### TO-BE
스크린샷 2023-09-24 오후 7 41 25

I haven't added tests for the parts where TODOs were left in the logic.

## PR Checklist

### DIFF
--- a/packages/react/src/experimental/Await.spec.tsx
+++ b/packages/react/src/experimental/Await.spec.tsx
@@ -2,7 +2,7 @@ import { act, render, screen, waitFor } from '@testing-library/react'
 import { vi } from 'vitest'
 import { ErrorBoundary, Suspense } from '..'
 import { ERROR_MESSAGE, FALLBACK, MS_100, TEXT, delay } from '../utils/toTest'
-import { awaitClient, useAwait } from '.'
+import { Await, awaitClient, useAwait } from '.'
 
 const key = ['key'] as const
 
@@ -94,5 +94,32 @@ describe('useAwait', () => {
     expect(screen.queryByText(FALLBACK)).toBeInTheDocument()
     act(() => vi.advanceTimersByTime(MS_100))
     await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument())
+  })
+})
+
+describe('Await component', () => {
+  it('should render child component with data from useAwait hook', async () => {
+    render(
+      <Suspense fallback="Loading...">
+        <Await options={{ key, fn: () => Promise.resolve(TEXT) }}>{({ data }) => <>{data}</>}</Await>
+      </Suspense>
+    )
+
+    expect(await screen.findByText(TEXT)).toBeInTheDocument()
+  })
+})
+
+describe('class AwaitClient clearError', () => {
+  beforeEach(() => awaitClient.reset())
+
+  it('clears promise & error for all cache without key', async () => {
+    try {
+      await awaitClient.suspend({ key, fn: () => Promise.reject(new Error(ERROR_MESSAGE)) })
+    } catch {
+      // Expected an error
+    }
+
+    awaitClient.clearError()
+    expect(awaitClient.getData([key])).toBeUndefined()
   })
 })


packages/react/src/experimental/Await.spec.tsx
```diff
@@ -96,3 +96,30 @@ describe('useAwait', () => {
...
+
+  it('clears promise & error for all cache without key', async () => {
+    try {
+      await awaitClient.suspend({ key, fn: () => Promise.reject(new Error(ERROR_MESSAGE)) })
+    } catch {
+      // Expected an error
+    }
+
+    awaitClient.clearError()
+    expect(awaitClient.getData([key])).toBeUndefined()
```
manudeli, line 123: This may be proper code

```suggestion
    expect(awaitClient.getData(key)).toBeUndefined()
```

packages/react/src/experimental/Await.spec.tsx
```diff
@@ -96,3 +96,30 @@ describe('useAwait', () => {
...
+      <Suspense fallback="Loading...">
+        <Await options={{ key, fn: () => Promise.resolve(TEXT) }}>{({ data }) => <>{data}</>}</Await>
+      </Suspense>
+    )
+
+    expect(await screen.findByText(TEXT)).toBeInTheDocument()
+  })
+})
+
+describe('class AwaitClient clearError', () => {
```
manudeli, line 112:
```suggestion
describe('awaitClient', () => {
```

packages/react/src/experimental/Await.spec.tsx
```diff
@@ -96,3 +96,30 @@ describe('useAwait', () => {
...
+  beforeEach(() => awaitClient.reset())
+
+  it('clears promise & error for all cache without key', async () => {
```
manudeli, line 115:
```suggestion
  it('should have clearError method to clear promise & error for all cache without key', async () => {
```

packages/react/src/experimental/Await.spec.tsx
```diff
@@ -96,3 +96,30 @@ describe('useAwait', () => {
     await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument())
   })
 })
+
+describe('Await component', () => {
```
manudeli, line 100: I want to change this line like this if we want to describe it as component more intuitively, after changing this file in this Pull Request, Let's change all files like this. Do you agree this? 

```suggestion
describe('<Await/>', () => {
```

packages/react/src/experimental/Await.spec.tsx
tooooo1: oh thx! Reviews related to text were reflected [bb1b64a](https://github.com/suspensive/react/pull/169/commits/bb1b64ae81a2d006274c8a4a05e2b48735b3d791)

packages/react/src/experimental/Await.spec.tsx
tooooo1: Certainly, using the JSX style like `<Await />` to represent components can be beneficial as it immediately signifies that it's a React component. However, as you've reviewed with me, [bb1b64a](https://github.com/suspensive/react/pull/169/commits/bb1b64ae81a2d006274c8a4a05e2b48735b3d791) most of the mentions were in the context of describe texts. Shifting to a JSX naming style might deviate from conventional practices. Also, standardization should be taken into consideration. Official documentation, tutorials, and community guidelines of testing libraries or frameworks generally don't recommend or mention using a JSX style for naming. Therefore, this style might differ from standard conventions or practices.

Have you ever seen this(<Await />) convention in the official testing documentation?

manudeli: When I see this [new official docs of react](https://react.dev/reference/react/components), I thought it's pretty intuitive way to express it as component than just using name. regardless of React's new official documentation, I think this way is a pretty good if the everyone can intuitively feel the code. then how about going our own way, even if we have nothing to refer to

image

tooooo1: I like it. It's definitely clear. I believe there won't be any issues as long as we maintain consistency moving forward. 

Should we reflect this now? Or should listen to Chanhyuk’s opinion?

manudeli: @okinawaa How do you think this?

okinawaa: Thank you for your valuable comments and reference At first, the "<Await/>" suggested by manudeli is the most impressive because **I think the test code is easy to read if it is similar to the way the component is actually used.** It's because we're developing it in the "<Await/>" way, not "<Await>"

tooooo1: b4bba05

packages/react/src/experimental/Await.spec.tsx
```diff
@@ -96,3 +96,30 @@ describe('useAwait', () => {
     await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument())
   })
 })
+
+describe('<Await>', () => {
```
manudeli, line 100: I understood @okinawaa's comment as using with /
```suggestion
describe('<Await/>', () => {
```

tooooo1: [1cdfd91](https://github.com/suspensive/react/pull/169/commits/1cdfd9186b20338de2c8c07153df62192d18a099)

manudeli (approved):
I left issue to change all describe for component as `<.../>`

PR #173, opened 2023-09-26
test(react): add isPlainObject case

tooooo1:
related with #62

# Overview
introduces a series of tests to ensure the correct behavior of the `isPlainObject` function.

AS-IS
image

TO-BE
image

## PR Checklist


manudeli (approved):
After merging this, I will devide isPlainObject as a new file

PR #177, opened 2023-09-27
fix(react): error boundary reset using render prop

codebutler:
# Overview

Fixes error:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'props')
    at reset (ErrorBoundary.tsx:85:10)
```

## PR Checklist

- [X] I have written documents and tests, if needed.

### DIFF
--- a/.changeset/yellow-clocks-deliver.md
+++ b/.changeset/yellow-clocks-deliver.md
@@ -2,4 +2,4 @@
 '@suspensive/react': patch
 ---
 
-fix error boundary reset using render prop
+fix(react): fix ErrorBoundary reset using render prop


manudeli (approved):
I missed this, thank you for fixing it👍

.changeset/yellow-clocks-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'@suspensive/react': patch
+---
+
+fix error boundary reset using render prop
```
manudeli, line 5:
```suggestion
fix(react): fix ErrorBoundary reset using render prop
```

PR #184, opened 2023-09-30
fix(react-query): correct peerDeps version

manudeli:
# Overview

Correct peerDeps version

## PR Checklist

### DIFF
--- a/packages/react-query/package.json
+++ b/packages/react-query/package.json
@@ -70,7 +70,7 @@
     "tsd": "^0.28.1"
   },
   "peerDependencies": {
-    "@suspensive/react": "^1.11.4-beta.0",
+    "@suspensive/react": "workspace:1.14.8",
     "@tanstack/react-query": "^4",
     "react": "^16.8 || ^17 || ^18"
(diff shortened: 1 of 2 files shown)


packages/react-query/package.json
```diff
@@ -70,7 +70,7 @@
     "tsd": "^0.28.1"
   },
   "peerDependencies": {
-    "@suspensive/react": "^1.11.4-beta.0",
+    "@suspensive/react": "workspace:1.14.8",
```
okinawaa, line 73: Can users can only use this version?!

okinawaa: I dont know well in this part
Isn't it common to use ^or an upper(>=) version?!

manudeli: Yes, I wanted to force users to use fixed versions of all suspensive packages
I thought that if we provided fixed versions of all suspensive packages as same package version, it would be easier for not only our maintainers and also for our users to manage versions as well.

okinawaa (approved):
Good!

Issue #179, opened 2023-09-27
[Feature]: Update documentation for all public apis


manudeli:
### **Package Scope**

- [x] Add to an existing package

Package name: @suspensive/react

### **Overview**

Update documentation for all public apis

PR #190, opened 2023-10-01
docs(react): update doc for ErrorBoundary

manudeli:
related with #179 

# Overview

I updated doc for ErrorBoundary

## PR Checklist

### DIFF
--- a/packages/react/src/ErrorBoundary.en.mdx
+++ b/packages/react/src/ErrorBoundary.en.mdx
@@ -3,7 +3,7 @@ sidebar_position: 2
 title: <ErrorBoundary/>
 ---
 
-Any rendering errors in children of this component can be handled.
+this component can handle any errors in children
 
 ### props.fallback
 
--- a/packages/react/src/ErrorBoundary.en.mdx
+++ b/packages/react/src/ErrorBoundary.en.mdx
@@ -7,7 +7,7 @@ Any rendering errors in children of this component can be handled.
 
 ### props.fallback
 
-If there is any thrown error in children of `<ErrorBoundary/>`, Error will be catched and then fallback will be rendered.
+If there is any thrown error in children of `<ErrorBoundary/>`, Error will be caught and then fallback will be rendered.
 
 ```tsx
 import { ErrorBoundary } from '@suspensive/react'
--- a/packages/react/src/ErrorBoundary.en.mdx
+++ b/packages/react/src/ErrorBoundary.en.mdx
@@ -120,7 +120,7 @@ const Example = () => (
 )
 ```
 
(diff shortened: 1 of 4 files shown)


packages/react/src/ErrorBoundary.en.mdx
```diff
@@ -3,16 +3,15 @@ sidebar_position: 2
 title: <ErrorBoundary/>
 ---
 
-This component provide a simple and reusable wrapper that you can use to wrap around your components. Any rendering errors in your components hierarchy can then be gracefully handled.
+Any rendering errors in children of this component can be handled.
```
okinawaa, line 6: Due to the `useErrorboundary`, errors other than rendering can be handled in the `Errorboundary`, is it okay not to specify that part?

packages/react/src/ErrorBoundary.en.mdx
```diff
@@ -3,16 +3,15 @@ sidebar_position: 2
...
 
-![Example banner](/gif/errorboundary-example.gif)
-
-## fallback
+### props.fallback
 
-If there is any thrown error in children, ErrorBoundary will catch it and then fallback will be rendered.
+If there is any thrown error in children of `<ErrorBoundary/>`, Error will be catched and then fallback will be rendered.
```
okinawaa, line 10:
```suggestion
If there is any thrown error in children of `<ErrorBoundary/>`, Error will be caught and then fallback will be rendered.
```

packages/react/src/ErrorBoundary.en.mdx
```diff
@@ -81,9 +120,18 @@ const Example = () => (
 )
 ```
 
-We can use useErrorBoundary.setError in ErrorBoundaryFallback to make ErrorBoundary aware of the Error without throw.
+We can use useErrorBoundary.setError in ErrorBoundaryFallback to make `<ErrorBoundary/>` aware of the Error without throw.
```
okinawaa, line 123: actually I usually use `setError` in children, not in ErrorBoundaryFallback😄

I think I'll use useErrorboundary.reset a lot in fallback component

packages/react/src/ErrorBoundary.en.mdx
manudeli: You caught wrong grammer! Thanks

packages/react/src/ErrorBoundary.en.mdx
manudeli: Oh! you're right, It's my mistake

packages/react/src/ErrorBoundary.en.mdx
manudeli: Wow... so good catch!👍

okinawaa (approved):

PR #196, opened 2023-10-02
feat(react): add all props type of all components

### LINKED ISSUE #194: [Feature]: All Props type to public apis
### **Package Scope**

manudeli:
fix #194 

# Overview

I added all props type of all components

## PR Checklist


packages/react/src/AsyncBoundary.tsx
```diff
@@ -1,10 +1,9 @@
-import { ComponentProps, ComponentRef, ComponentType, SuspenseProps, forwardRef } from 'react'
-import { ErrorBoundary } from './ErrorBoundary'
-import { Suspense } from './Suspense'
+import { ComponentProps, ComponentRef, ComponentType, forwardRef } from 'react'
+import { ErrorBoundary, ErrorBoundaryProps } from './ErrorBoundary'
```
minsoo-web, line 2: How about use `import type` ? or `type { FooProps } from "somewhere"`
Or add tsconfig `importsNotUsedAsValues` to `remove` !

ref: https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues
https://cloudless.blog/post/import%20type%EA%B3%BC%20import

https://stackoverflow.com/questions/50441093/is-there-a-point-to-doing-import-type-rather-than-import-with-flow

it's eslint config exist
```json
"@typescript-eslint/consistent-type-imports": "error",
```

manudeli: I resolve this in https://github.com/suspensive/react/pull/219

minsoo-web (approved):
LGTM 👍

PR #198, opened 2023-10-02
docs(react): update doc for Delay

manudeli:
# Overview

I updated doc for Delay

## PR Checklist


manudeli:
@okinawaa, I merge first, let's make Pull Reqeust if we need to change

PR #207, opened 2023-10-05
docs(react): update tips as admonitions

manudeli:
related with #179 

# Overview

I updated tips as admonitions

@okinawaa, when you compare change, I think the easiest way to compare the differences is to open two screens, one with dev server on one side and the other with production suspensive.org. I would appreciate it if you could make that comparison.

## PR Checklist


manudeli:
@okinawaa There is no big change, just style change so I merge to reduce your effort. If we have other issue, let's make PR after

Issue #200, opened 2023-10-03
[Feature]: Remove experimental mark of useErrorBoundary


manudeli:
### **Package Scope**

- [x] Add to an existing package

Package name: @suspensive/react

### **Overview**

Remove unnecessary field of useErrorBoundary

PR #216, opened 2023-10-11
feat(react): add useErrorBoundaryFallbackProps

manudeli:
related with #200 

# Overview

I added **useErrorBoundaryFallbackProps** to make **useErrorBoundary** don't have role in fallback of `<ErrorBoundary/>`

## Cause of adding **useErrorBoundaryFallbackProps**
1. TypeScriptly, existing useErrorBoundary().isError can be true in **children of `<ErrorBoundary/>`**, but actually runtimely it's not.
    ```tsx
    const Example = withErrorBoundary(() => {
      errorBoundary.isError // there cannot be error, but isError can be true TypeScriptly
    }, {
      fallback: () => <></>
    })
    ```
2. TypeScriptly, existing useErrorBoundary().isError can be false in **fallback of `<ErrorBoundary/>`**, but actually runtimely there must have error in fallback of `<ErrorBoundary/>`
    ```tsx
    const Example = withErrorBoundary(() => {
      
    }, {
      fallback: function ErrorBoundaryFallback() => {
        errorBoundary.isError // there must be error, but isError can be false TypeScriptly
        return <></>
      }
    })
    ```
3. There could be useErrorBoundary().reset in children of `<ErrorBoundary/>`, but it didn't work in children because there is no error to reset in children of `<ErrorBoundary/>`.
    ```tsx
    const Example = withErrorBoundary(() => {
      errorBoundary.reset // there is reset method but this is not working for children of `<ErrorBoundary/>`
    }, {
      fallback: () => <></>
    })
    ```

## Solution
I think both children / fallback of `<ErrorBoundary/>` should have separate hook for each section. so I made useErrorBoundaryFallbackProps separate from useErrorBoundary

- useErrorBoundary(hook for children of `<ErrorBoundary/>`)
  - everytime no error
  - no reset function
  - can setError
- useErrorBoundaryFallbackProps(hook for fallback of `<ErrorBoundary/>`)
  - everytime have error
  - reset function working

## PR Checklist


packages/react/src/ErrorBoundary.en.mdx
```diff
@@ -48,6 +48,38 @@ const ErrorAfter4s = () => {
...
+    <>
+      <button onClick={props.reset} />
+      {props.error.message}
+    </>
+  )
+}
+
+const ErrorBoundaryFallback = () => <Nested />
+const Example = () => (
+  <ErrorBoundary fallback={ErrorBoundaryFallback}>
```
minsoo-web, line 75:
```suggestion
const Example = () => (
  <ErrorBoundary fallback={Nested}>
```

What about this?
I think `ErrorBoundaryFallback` component is not necessary
or, Nested component rename to ErrorBoundaryFallback

minsoo-web, line 75: How about add onReset prop ??

User who show this example or document first time, just draged whole example and paste their code. and curious "why my query is not reseted and don't query again?"

packages/react/src/ErrorBoundary.en.mdx
```diff
@@ -48,6 +48,38 @@ const ErrorAfter4s = () => {
...
+
+```tsx
+import { ErrorBoundary, useErrorBoundaryFallbackProps } from '@suspensive/react'
+
+const Nested = () => {
+  const props = useErrorBoundaryFallbackProps()
+
+  return (
+    <>
+      <button onClick={props.reset} />
```
minsoo-web, line 67:
```suggestion
      <button onClick={props.reset}>click to reset error!</button>
```

how about like this?

packages/react/src/ErrorBoundary.en.mdx
```diff
@@ -48,6 +48,38 @@ const ErrorAfter4s = () => {
 
 ![Example banner](/gif/errorboundary-example.gif)
 
+:::tip
+
+## Use `<ErrorBoundary/>` fallback props without prop drilling
+
+### useErrorBoundaryFallbackProps
```
minsoo-web, line 55: I think we need to seperate component's doc and hook's doc.

I think you intended combine this hook with component document, because it is related to this component.
But I think it is more convenient to seperated component document  with hook document

reference is below
image

Chakra's intend is like that,
> If you need to use hook, check this [link](https://chakra-ui.com/docs/components/checkbox/usage#hooks)

image

But our document is mixed.

minsoo-web (commented):
I think this hook is very convenient, but how about provide props type guide like this? 

```
import { ErrorBoundaryFallbackProps } from '@suspensive/react'

const Nested = ({ reset, error }: ErrorBoundaryFallbackProps) => {
  return (
    <>
      <button onClick={reset}>reset</button>
      {error.message}
    </>
  )
}
```

This is exactly same, but can't avoid props drilling.
So, my suggestion is, before provide example that can avoid props drilling case, first show simple example.

packages/react/src/ErrorBoundary.en.mdx
minsoo-web: Oh you intend to props drilling case... I miss understand sorry

packages/react/src/ErrorBoundary.tsx
```diff
@@ -20,11 +20,11 @@ import { PropsWithoutChildren } from './types'
 import { hasResetKeysChanged } from './utils'
 import { assert } from './utils'
```
minsoo-web, line 21: I think this is combined with one line.

What about add vscode's config like this

```
"editor.codeActionsOnSave": {
    "source.fixAll.eslint": true,
    "source.organizeImports": true
  },
```
> .vscode/settings.json

packages/react/src/ErrorBoundary.ko.mdx
```diff
@@ -48,6 +48,38 @@ const ErrorAfter4s = () => {
 
 ![Example banner](/gif/errorboundary-example.gif)
 
+:::tip
+
+## `<ErrorBoundary/>` fallback props을 prop drilling 없이 사용하기
+
+### useErrorBoundaryFallbackProps
+
+ErrorBoundaryFallback에서 useErrorBoundaryFallbackProps().reset/error을 사용해 prop을 직접사용하지 않고도 `<ErrorBoundary/>`을 reset하고 error를 사용할 수 있습니다.
```
SEOKKAMONI, line 57:
```suggestion
ErrorBoundaryFallback에서 useErrorBoundaryFallbackProps().reset/error을 사용해 prop을 직접 사용하지 않고도 `<ErrorBoundary/>`을 reset하고 error를 사용할 수 있습니다.
```

packages/react/src/ErrorBoundary.ko.mdx
```diff
@@ -48,6 +48,38 @@ const ErrorAfter4s = () => {
 
 ![Example banner](/gif/errorboundary-example.gif)
 
+:::tip
+
+## `<ErrorBoundary/>` fallback props을 prop drilling 없이 사용하기
```
SEOKKAMONI, line 53:
```suggestion
## `<ErrorBoundary/>` fallback props을 props drilling 없이 사용하기
```

how about like this?

packages/react/src/ErrorBoundary.ko.mdx
```diff
@@ -48,6 +48,38 @@ const ErrorAfter4s = () => {
...
+    <Error />
+  </ErrorBoundary>
+)
+```
+
+:::
+
 ### props.resetKeys
 
 `<ErrorBoundary/>`의 fallback 외부에 있는 컴포넌트가 `<ErrorBoundary/>`를 reset하려면 resetKeys배열에 resetKey를 할당하면 됩니다. resetKeys는 배열의 하나 이상의 요소가 변경된 경우에만 작동합니다. useEffect의 종속성 배열이 작동하는 방식과 같이 resetKeys로 매 렌더링마다 새 배열을 주입하는 것을 걱정할 필요도 없습니다.
```
SEOKKAMONI, line 85:
```suggestion
`<ErrorBoundary/>`의 fallback 외부에 있는 컴포넌트가 `<ErrorBoundary/>`를 reset하려면 resetKeys 배열에 resetKey를 할당하면 됩니다. resetKeys는 배열의 하나 이상의 요소가 변경된 경우에만 작동합니다. useEffect의 종속성 배열이 작동하는 방식과 같이 resetKeys로 매 렌더링마다 새 배열을 주입하는 것을 걱정할 필요도 없습니다.
```
how about like this?

packages/react/src/ErrorBoundary.ko.mdx
```diff
@@ -123,25 +155,9 @@ useErrorBoundary는 실험 기능이므로 이 인터페이스는 변경될 수
...
-const Example = () => (
-  <ErrorBoundary fallback={ErrorBoundaryFallback}>
-    <Error />
-  </ErrorBoundary>
-)
-```
+### useErrorBoundary().setError
 
-`<ErrorBoundary/>`의 children에서 useErrorBoundary.setError을 사용해 throw 없이도 `<ErrorBoundary/>`에서 Error를 알도록 할 수 있습니다.
+`<ErrorBoundary/>`의 children에서 useErrorBoundary().setError을 사용해 throw 없이도 `<ErrorBoundary/>`에서 Error를 알도록 할 수 있습니다.
```
SEOKKAMONI, line 160:
```suggestion
`<ErrorBoundary/>`의 children에서 useErrorBoundary().setError을 사용해 throw 없이도 `<ErrorBoundary/>`에서 Error를알게 할 수 있습니다.
```

how about like this?

packages/react/src/ErrorBoundary.ko.mdx
manudeli: Thanks, but I think `prop drilling` is right (https://react.dev/learn/passing-data-deeply-with-context#the-problem-with-passing-props)

manudeli:
@SEOKKAMONI Thanks👍

packages/react/src/ErrorBoundary.tsx
manudeli: I resolve this by eslint rule, not vscode setting in https://github.com/suspensive/react/pull/219

packages/react/src/ErrorBoundary.en.mdx
manudeli: I made https://github.com/suspensive/react/issues/220 to resolve it. Let's talk about it after this Pull Request

packages/react/src/ErrorBoundary.en.mdx
manudeli: I updated it in [1e748f8](https://github.com/suspensive/react/pull/216/commits/1e748f837c19d8e3cd4214e363fece8e62ddfc6e)

Almost cases we use text as `Try again` to reset ErrorBoundary

packages/react/src/ErrorBoundary.en.mdx
manudeli: > and curious "why my query is not reseted and don't query again?"

I guess `query` you mentioned is for @tanstack/react-query. but because @suspensive/react is not only for @tanstack/react-query, I think we should not to describe this example with @tanstack/react-query

manudeli:
> I think this hook is very convenient, but how about provide props type guide like this? 
> 
> ```tsx
> import { ErrorBoundaryFallbackProps } from '@suspensive/react'
> 
> const Nested = ({ reset, error }: ErrorBoundaryFallbackProps) => {
>   return (
>     <>
>       <button onClick={reset}>reset</button>
>       {error.message}
>     </>
>   )
> }
> ```
> 
> This is exactly same, but can't avoid props drilling.
> So, my suggestion is, before provide example that can avoid props drilling case, first show simple example.
> 
> ---
> 
> And How about add Errorboundary component's visualization page and link it?
> 
> hundred words are worth a glance. In korea `백문이불여일견`

@minsoo-web could you add suggestion code or Pull Request to this branch please? I want to know what your intend exactly

packages/react/src/ErrorBoundary.en.mdx
minsoo-web: I agree, thanks!

minsoo-web:
> @minsoo-web could you add suggestion code or Pull Request to this branch please? I want to know what your intend exactly

I made PR for this case #221 
Please review this PR ! 🙇‍♂️

packages/react/src/ErrorBoundary.en.mdx
```diff
@@ -48,6 +48,38 @@ const ErrorAfter4s = () => {
...
+const Nested = () => {
+  const props = useErrorBoundaryFallbackProps()
+
+  return (
+    <>
+      <button onClick={props.reset}>Try again</button>
+      {props.error.message}
+    </>
+  )
+}
```
SEOKKAMONI, line 71:
```suggestion
```tsx
import { ErrorBoundary, useErrorBoundaryFallbackProps } from '@suspensive/react'

const Nested = () => {
  const { reset, error } = useErrorBoundaryFallbackProps()

  return (
    <>
      <button onClick={reset}>Try again</button>
      {error.message}
    </>
  )
}
```

packages/react/src/ErrorBoundary.ko.mdx
```diff
@@ -48,9 +48,41 @@ const ErrorAfter4s = () => {
...
+const Nested = () => {
+  const props = useErrorBoundaryFallbackProps()
+
+  return (
+    <>
+      <button onClick={props.reset}>Try again</button>
+      {props.error.message}
+    </>
+  )
+}
```
SEOKKAMONI, line 71:
```suggestion
  const { reset, error } = useErrorBoundaryFallbackProps()

  return (
    <>
      <button onClick={reset}>Try again</button>
      {error.message}
    </>
  )
}
```

SEOKKAMONI (commented):
I referenced to #221, how about this for document commonization?

minsoo-web (approved):
LGTM 👍

packages/react/src/ErrorBoundary.en.mdx
```diff
@@ -48,6 +48,68 @@ const ErrorAfter4s = () => {
...
+    <button onClick={reset}>reset</button>
+    {error.message}
+  </>
+)
+
+const Example = () => (
+  <ErrorBoundary fallback={ErrorBoundaryFallback}>
+    <ErrorAfter4s />
+  </ErrorBoundary>
+)
```
okinawaa, line 73: If it's for the purpose of providing typing

Wouldn't it be better for the developer experience to define not only the type of props but also the return?

```suggestion
import type { ErrorBoundaryFallbackType } from '@suspensive/react'

const ErrorBoundaryFallback: ErrorBoundaryFallbackType = ({ reset, error }) => (
  <>  // The type of return value is also certain
    <button onClick={reset}>reset</button>
    {error.message}
  </>
)

const Example = () => (
  <ErrorBoundary fallback={ErrorBoundaryFallback}> 
    <ErrorAfter4s />
  </ErrorBoundary>
)
```

I inspried by nextjs GetServerSideProps

okinawaa (approved):

packages/react/src/ErrorBoundary.en.mdx
manudeli: > Wouldn't it be better for the developer experience to define not only the type of props but also the return?

I think that's a good idea too. However, even if we do not provide the return type, TypeScript will create a type error if it does not match the fallback type of ErrorBoundary.

I think we need to consider whether providing a return type is right choice. Because the return type may be different from the actual thing returned ([this video](https://www.youtube.com/watch?v=I6V2FkW1ozQ&ab_channel=Theo-t3%E2%80%A4gg) explain it well). If it is as essential as ErrorBoundaryFallbackProps, it would be a good idea to add it as a minor version feature.

Let's consider it after merging this

PR #239, opened 2023-10-19
feat(hook/utils): add useIsomorphicLayoutEffect

minsoo-web:
# Overview

To distinguish whether it is a client or server environment from a hook or component using `useEffect`

I will change the `useEffect` used in places where the client environment and the server environment should be distinguished to the `useIsomorphicLayoutEffect`.

## PR Checklist

1. I read the [Contributing Guide](https://github.com/toss/slash/blob/main/.github/CONTRIBUTING.md)


packages/react/src/utils/isServer.ts
```diff
@@ -0,0 +1,3 @@
+export const isServer = () => {
+  return typeof window === 'undefined' && typeof global !== 'undefined'
```
manudeli, line 2: I don't know why we should check global. could you explain it please?

No global checking libraries references
- usehoos-ts: https://usehooks-ts.com/react-hook/use-isomorphic-layout-effect
- react-use: https://github.com/streamich/react-use/blob/master/src/useIsomorphicLayoutEffect.ts

packages/react/src/utils/index.ts
```diff
@@ -1,3 +1,5 @@
+export { assert } from './assert'
 export { hasResetKeysChanged } from './hasResetKeysChanged'
 export { hashKey } from './hashKey'
-export { assert } from './assert'
+export { isClient } from './isClient'
+export { isServer } from './isServer'
```
manudeli, line 5: I think we don't need to make unnecessary function signature now. if there is needs to make it after, Let's make it then.

```suggestion
```

packages/react/src/hooks/useIsomorphicLayoutEffect.ts
```diff
@@ -0,0 +1,6 @@
+import { useEffect, useLayoutEffect } from 'react'
+import { isClient } from '../utils/isClient'
+
+const isClientEnv = isClient()
+
+export const useIsomorphicLayoutEffect = isClientEnv ? useLayoutEffect : useEffect
```
manudeli, line 6: How about simply now like below?

check window only
```suggestion
import { useEffect, useLayoutEffect } from 'react'

export const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? useLayoutEffect : useEffect
```

or


check document only
```suggestion
import { useEffect, useLayoutEffect } from 'react'

export const useIsomorphicLayoutEffect = typeof document !== 'undefined' ? useLayoutEffect : useEffect
```

packages/react/src/utils/isServer.ts
minsoo-web: I think, if in node env or server env, `this` is global so this check is equal to `typeof window === 'undefined'`

packages/react/src/hooks/useIsomorphicLayoutEffect.ts
minsoo-web: What is different `isClient` function call case and `typeof document !== 'undefined'` case?
I don't know exactly about your intension.

manudeli: Then how about this. I thought we don't need to make extra code to make `isClient` as function yet.

1. no reusing `isClient` yet
2. function `isClient` will be contained in bundle
3. In our case, We don't have to recall it. so we don't have to make `isClient` as funciton

```suggestion
import { useEffect, useLayoutEffect } from 'react'
const isClient = typeof window !== 'undefined'
export const useIsomorphicLayoutEffect = isClient ? useLayoutEffect : useEffect
```

if we want to use as extra module, I think below is enough, not `isClient` as function

```ts
// src/utils/isClient.ts
export const isClient = typeof window !== 'undefined'
```

```suggestion
import { useEffect, useLayoutEffect } from 'react'
import { isClient } from '../utils' // using isClient as just boolean is okay

export const useIsomorphicLayoutEffect = isClient ? useLayoutEffect : useEffect
```

packages/react/src/utils/isServer.ts
manudeli: I can accept semantical cause. but I think if we don't check global in this case, we can get better performance. because we are making most fundamental library so this isClient function will be recalled many time. so I don't want to give up our better performance if we can

I asked this to slash library too. https://github.com/toss/slash/issues/352

Every libraries I checked don't check global to know it is server or client. so I'm very curious your and slash's intention

minsoo-web: I agree. this function is called every time in bundler position. 
So this expression called inline, rather than export.

minsoo-web: [chore: delete isClient, isServer function](https://github.com/suspensive/react/pull/239/commits/6a2514d5411688131230be159edb11da6fd5ffc4)

I delete it !! many thanks 🚀

manudeli:
I think you need to use corepack with pnpm to meet your pnpm version as packageManager field of package.json. check this doc https://pnpm.io/installation#using-corepack

This is good for install it
https://github.com/TanStack/query/blob/main/package.json#L12

packages/react/src/utils/isServer.ts
manudeli: Thanks for accepting my opinion, Let's improve this if we need to improve it after

manudeli:
> I think you need to use corepack with pnpm to meet your pnpm version as packageManager field of package.json. check this doc https://pnpm.io/installation#using-corepack
> 
> This is also good for install script https://github.com/TanStack/query/blob/main/package.json#L12

I resolved this issue in [80cec91](https://github.com/suspensive/react/pull/239/commits/80cec91e183e5fe28c9fbfe03f4d6bdb1777af4f) please use corepack please. I think we need to add this in CONTRIBUTING.md

manudeli (approved):

PR #246, opened 2023-10-20
fix(react): rename useIsMount & update useEffect to useIsomorphicLayoutEffect

minsoo-web:
# Overview

## I update useIsMount hook's name to useIsClient

Although I need to think more about the name, I named this hook because it is not an externally exported package, but an embedded hook used to distinguish whether it is a client environment or not.

## I update useIsClient hook inside of Suspense.CSROnly from useEffect to useIsomorphicLayoutEffect

Although it is used differently depending on where it is used, `useEffect` is called at the mount stage, but `useLayoutEffect` is executed synchronously immediately after rendering, so it is possible to check whether it is in a client state or not first, so this hook is used.

## PR Checklist


packages/react/src/Suspense.tsx
```diff
@@ -21,10 +21,10 @@ if (process.env.NODE_ENV !== 'production') {
   DefaultSuspense.displayName = 'Suspense'
 }
 const CSROnlySuspense = (props: SuspenseProps) => {
-  const isMounted = useIsMounted()
+  const isClient = useIsCSROnly()
```
manudeli, line 24:
```suggestion
  const isClient = useIsClient()
```

Can we change the name of this hook to useIsClient? If this hook is likely to be reused elsewhere in the future, I don't think it will only be used in CSROnly. In that case, I think a better name would be useIsClient rather than CSROnly. Looking at the code, the return value is also the variable name isClient, so I think useIsClient would be a better name. Because the name CSROnly also means Client Side Render Only, which can contain isClient.

okinawaa: Im down

Then our Suspense 
Should it be `Suspense.Client`, not `Suspense.CSROnly`?

manudeli: @okinawaa I think`<Suspense.CSROnly/>` using `useIsClient` is best case.
then we will not go through `BREAKING CHANGE` and also naming this component as `<Suspense.Client/>` is not more clear to express `client side render only` than `<Suspense.CSROnly/>`.

minsoo-web: @manudeli 
I agree with you. 

CSROnly naming will keep, and useIsCSROnly naming is will changed to useIsClient. I will do that

manudeli: Thanks for accepting my opinion

manudeli (approved):

manudeli:
@okinawaa @minsoo-web This is test using visualization

### AS-IS useEffect in CSROnly
If no useIsomorphicLayoutEffect, We can check red loading... fallback on every time
chrome-capture-2023-9-23

### TO-BE useIsomorphicLayoutEffect in CSROnly
no red loading... fallback on every time. issue resolved
chrome-capture-2023-9-23 (1)

```tsx
// src/app/react/CSROnly1
'use client'

import { Suspense } from '@suspensive/react'
import Link from 'next/link'
import React from 'react'

export default function Page() {
  return (
    <div>
      <Suspense.CSROnly fallback={<div style={{ backgroundColor: 'red' }}>loading...</div>}>
        <Custom/>
      </Suspense.CSROnly>
    </div>
  )
}

const Custom = () => {
  return (
    <div>
      CSROnly1
      <Link href={'/react/CSROnly2'}>to CSROnly2</Link>
    </div>
  )
}

```


```tsx
// src/app/react/CSROnly2
'use client'

import { Suspense } from '@suspensive/react'
import Link from 'next/link'
import React from 'react'

export default function Page() {
  return (
    <div>
      <Suspense.CSROnly fallback={<div style={{ backgroundColor: 'red' }}>loading...</div>}>
        <Custom/>
      </Suspense.CSROnly>
    </div>
  )
}

const Custom = () => {
  return (
    <div>
      CSROnly2
      <Link href={'/react/CSROnly1'}>to CSROnly1</Link>
    </div>
  )
}

```

manudeli:
I think this is cool work! big thanks to @minsoo-web making issue and resolving it 👍👍👍
@okinawaa This is an important improvement, so I hope you approve and merge this Pull Request if there are no further issues.

okinawaa (approved):
cool 👍

PR #256, opened 2023-10-23
refactor(test-utils): add package for test utils

manudeli:
# Overview

I add new dev package @suspensive/test-utils to gather test-utils
resolve https://github.com/suspensive/react/pull/234#discussion_r1365336296

## PR Checklist


configs/test-utils/package.json
```diff
@@ -0,0 +1,50 @@
...
+        "default": "./dist/index.cjs"
+      }
+    },
+    "./package.json": "./package.json"
+  },
+  "main": "dist/index.cjs",
+  "types": "dist/index.d.ts",
+  "scripts": {
+    "build": "tsup",
+    "build:watch": "tsup --watch",
```
minsoo-web, line 33: Is this build config required? I think there is no need to build this files and need to just export pure ts file. isn't it?

minsoo-web (commented):
Cool change! 👍

configs/test-utils/package.json
manudeli: Actually, I tried it before, but it didn't work properly so I gave up. However, when I looked again, I should have used index.tsx as main, but the problem occurred because I used index.ts.

So, to avoid confusion in the first place, we unified them into index.ts.
Thank you for giving me motivation to try again

minsoo-web (approved):
LGTM 👍 !!

PR #262, opened 2023-10-25
chore(pnpm): force corepack by preinstall

### LINKED ISSUE #243: [Feature] corepack enable in CONTRIBUTING.md
> I think you need to use corepack with pnpm to meet your pnpm version as packageManager field of package.json. check this doc https://pnpm.io/installation#using-corepack
> 
> This is also good for install script https://github.com/TanStack/query/blob/main/package.json#L12

I resolved this issue in [80cec91](https://github.com/suspensive/react/pull/239/commits/80cec91e183e5fe28c9fbfe03f4d6bdb1777af4f) please use corepack please. I think we need to add this in CONTRIBUTING.md

_Originally posted by @manudeli in https://github.com/suspensive/react/issues/239#issuecomment-1772601247_

manudeli:
fix #243 

# Overview

If we force [corepack, we don't need to notice what version we need to use for pnpm](https://pnpm.io/installation#using-corepack).
So I use preinstall of package.json and update CONTRIBUTING.md

## PR Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "lint:pack": "packlint sort -R",
     "lint:pub": "turbo run lint:pub",
     "postinstall": "husky install",
+    "preinstall": "corepack enable",
     "prepack": "turbo run prepack",
     "test": "turbo run test",
     "test:tsd": "turbo run test:tsd",
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,7 +42,8 @@ In general, things we find useful when reviewing suggestions are:
     ```shell
     git clone git@github.com:suspensive/react.git
     ```
-2. Install relative packages. [We use pnpm v8. Install it please if need](https://pnpm.io/installation)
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -39,6 +39,7 @@
     "lint:pack": "packlint sort -R",
     "lint:pub": "turbo run lint:pub",
     "postinstall": "husky install",
+    "preinstall": "corepack enable",
```
minsoo-web, line 42: do we need to postinstall command ? it is listened git history

minsoo-web (approved):

package.json
manudeli: More explanation is needed

minsoo-web: If I know exactly right, I understand that this command is only necessary if we don't manage the `.husky` folder with git. 
We are managing husky with git, so I think it's a command we don't need.

---

I tested it in detail, and it didn't work without `.husky/_` folders, and `.husky/_` is a folder registered in gitignore, so I think I need the husky install command every time. Is it bad to commit `.husky/_` to git?

PR #264, opened 2023-10-26
perf(react): remove unnecessary useSetTimeout's callback calling

### LINKED ISSUE #263: [Bug]: Issues with the "useSetTimeout" behavior
### Package Scope

@suspensive/react

### Bug description

I found a behavioral issue with`useSetTimeout`, but that may or may not be a bug, depending on your actual intent. 

Currently, `useSetTimeout` is written as follows, which causes `useSetTimeout` to run repeatedly every time the component is re-rendered. Is this behavior intended?

```tsx
const [bool, setBool] = useState(false);

useSetTimeout(() => {
  console.log("If it is re-rendered, it will run again.");
}, 500);

return () => {
  <button onClick={() => setBool(!bool)}>rerender</button>
}
```

### Expected behavior

I think `useSetTimeout` should only run once after the initial render, not repeatedly every time the component is re-rendered. It would be confusing if it ran every time it was re-rendered.

### To Reproduce

_No response_

### Possible Solution

we need to memoize the callback function we pass as an argument and keep a reference to it.

```tsx
export const useSetTimeout = (fn: (...args: []) => void, delay: number) => {
  const callbackFunc = useRef(fn);
  
  useEffect(() => {
    callbackFunc.current = fn;
  }, [fn]);

  useEffect(() => {
    const timeout = setTimeout(callbackFunc.current, delay)
    return () => clearTimeout(timeout)
  }, [delay])
}
```

Alternatively, a custom hook like [usePreservedCallback in toss/slash](https://github.com/toss/slash/blob/main/packages/react/react/src/hooks/usePreservedCallback.ts) would be a good way to go. 

### etc.

https://github.com/streamich/react-use/blob/master/src/useTimeoutFn.ts

https://github.com/juliencrn/usehooks-ts/blob/master/packages/usehooks-ts/src/useTimeout/useTimeout.ts

Hooks and many related libraries are already written that way.

manudeli:
fix #263 

# Overview

@ssi02014 Thanks for reporting important issue
I remove unnecessary useSetTimeout's callback calling. Could you review this Pull Request?

## PR Checklist


configs/test-utils/src/ThrowError.tsx
```diff
@@ -1,11 +1,21 @@
 import type { PropsWithChildren } from 'react'
-import { useEffect, useState } from 'react'
+import { useEffect, useLayoutEffect, useRef, useState } from 'react'
+
+const isClient = typeof window !== 'undefined'
+export const useIsomorphicLayoutEffect = isClient ? useLayoutEffect : useEffect
+
+export const useTimeout = (fn: () => void, ms: number) => {
```
manudeli, line 7: I renamed useSetTimeout to useTimeout, because we use just Timeout not setTimeout

configs/test-utils/src/ThrowError.tsx
```diff
@@ -1,11 +1,21 @@
...
+  const fnRef = useRef(fn)
+
+  useIsomorphicLayoutEffect(() => {
```
manudeli, line 10: I used useIsomorphicLayoutEffect in this time, thanks to @minsoo-web 

I referred to [this code of usehooks-ts recommended](https://github.com/juliencrn/usehooks-ts/blob/master/packages/usehooks-ts/src/useTimeout/useTimeout.ts) by @ssi02014 a lot. thanks!👍

configs/test-utils/src/ThrowError.tsx
```diff
@@ -1,11 +1,21 @@
...
 
-const useSetTimeout = (fn: (...args: []) => void, delay: number) =>
   useEffect(() => {
-    const timeout = setTimeout(fn, delay)
-    return () => clearTimeout(timeout)
-  }, [fn, delay])
+    const id = setTimeout(() => fnRef.current(), ms)
+    return () => clearTimeout(id)
+  }, [ms])
+}
```
manudeli, line 18: remove unncessary fn calling

manudeli (commented):
I left comments as my intention

manudeli:
@ssi02014 If you okay, please approve this Pull Request

manudeli:
Ah, we need to add test case to check there is no unncessary useTimeout's fn calling on rerendering. I will add it to prove this fixing will work

ssi02014:
@manudeli 

Yes, I agree. The old `useSetTimeout` would fail the test code below. But with the new `useTimeout`, the test passes.

(The test code below is just a reference example, written in `jest` instead of `vitest`. Also, `rerender` in `renderHook` didn't work well with the test, so I redefined the component and utilized `rerender` in `render`).

```tsx
const delayTime = 1000;

const TestComponent = () => {
  const [number, setNumber] = useState(0);

  useTimeout(() => {
    setNumber(number + 1);
  }, delayTime);

  return <div>{number}</div>;
};
```

```tsx
it('should not be re-called even if the component is re-rendered', () => {
  const { rerender } = renderSetup(<TestComponent />);

  expect(screen.getByText('0')).toBeInTheDocument();

  act(() => jest.advanceTimersByTime(delayTime));
  expect(screen.getByText('1')).toBeInTheDocument();

  rerender(<TestComponent />);

  act(() => jest.advanceTimersByTime(delayTime))
  expect(screen.getByText('1')).toBeInTheDocument();
});
```

manudeli:
> @manudeli 
> 
> Yes, I agree. The old `useSetTimeout` would fail the test code below. But with the new `useTimeout`, the test passes.
> 
> (The test code below is just a reference example, written in `jest` instead of `vitest`. Also, `rerender` in `renderHook` didn't work well with the test, so I redefined the component and utilized `rerender` in `render`).
> 
> ```tsx
> const delayTime = 1000;
> 
> const TestComponent = () => {
>   const [number, setNumber] = useState(0);
> 
>   useTimeout(() => {
>     setNumber(number + 1);
>   }, delayTime);
> 
>   return <div>{number}</div>;
> };
> ```
> 
> ```tsx
> it('should not be re-called even if the component is re-rendered', () => {
>   const { rerender } = renderSetup(<TestComponent />);
> 
>   expect(screen.getByText('0')).toBeInTheDocument();
> 
>   act(() => jest.advanceTimersByTime(delayTime));
>   expect(screen.getByText('1')).toBeInTheDocument();
> 
>   rerender(<TestComponent />);
> 
>   act(() => jest.advanceTimersByTime(delayTime))
>   expect(screen.getByText('1')).toBeInTheDocument();
> });
> ```

Oh, cool! If you okay, could you make Pull Request to this branch too?

ssi02014:
@manudeli  
Yes! If the current Pull Request reflects, I'll work on the test code to create a new Pull Request! is that okay?

manudeli:
> @manudeli  
> Yes! If the current Pull Request reflects, I'll work on the test code to create a new Pull Request! is that okay?
> 
> Oh sorry I didn't catch your intent, can I just keep the current pull request and work on it?

After merging this, new Pull Request to add test case also okay! Thanks

configs/test-utils/src/ThrowError.tsx
minsoo-web: How about we use ms package for this? https://www.npmjs.com/package/ms
It is really good for dx.

configs/test-utils/src/ThrowError.tsx
```diff
@@ -14,7 +24,7 @@ export const ThrowError = ({ message, after = 0, children }: ThrowErrorProps) =>
   if (isNeedError) {
     throw new Error(message)
   }
-  useSetTimeout(() => setIsNeedError(true), after)
+  useTimeout(() => setIsNeedError(true), after)
   return <>{children}</>
```
minsoo-web, line 28: is this need for wrapping Fragment component?

configs/test-utils/src/ThrowError.tsx
```diff
@@ -24,7 +34,7 @@ export const ThrowNull = ({ after, children }: ThrowNullProps) => {
   if (isNeedError) {
     throw null
   }
-  useSetTimeout(() => setIsNeedError(true), after)
+  useTimeout(() => setIsNeedError(true), after)
   return <>{children}</>
```
minsoo-web, line 38: is this need for wrapping Fragment component? too!

packages/react/src/hooks/useTimeout.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useRef } from 'react'
+import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect'
+
+export const useTimeout = (fn: () => void, ms: number) => {
```
minsoo-web, line 4: this hook is needable to export ? why don't we import from test-util? it is duplicated

minsoo-web (commented):
Cool!

configs/test-utils/src/ThrowError.tsx
manudeli: I understood that you meant to use `ms` inside `suspensive`. But wouldn't it be more useful to use it like the code below? If `ms` is unnecessary, it can prevent unnecessary dependencies from being installed and unnecessary calculations to be performed. Also, the unit of time desired by the Delay component is quite small. In most cases, I expect to use `ms prop` for Delay of less than 2000ms, but I am wondering whether it would be a good idea to use `ms` internally in this situation. If you need to include it, I think it would be a good idea to create `ms` yourself so that you can include only the necessary parts.

### Delay using with ms is okay
```tsx
import ms from 'ms'

<Delay ms={ms('2 days')}>
  {/* ... */}
</Delay>
```

### Delay have to make BREAKING CHANGE if we use `ms` internally
```tsx
import ms from 'ms'

<Delay time={'2 days'}> // this ms prop have to make BREAKING CHANGE. because '2 days' is not just ms prop
  {/* ... */}
</Delay>
```

configs/test-utils/src/ThrowError.tsx
manudeli: image

configs/test-utils/src/ThrowError.tsx
manudeli: https://github.com/suspensive/react/pull/264#discussion_r1373686578

packages/react/src/hooks/useTimeout.ts
manudeli: @minsoo-web ssi02014 commented why same with exactly my intention.
also @suspensive/test-utils is just devDependency of @suspensive/react. so we need to import from internal module.

If you know alternative way, make Pull Request or add Issue please.

configs/test-utils/src/ThrowError.tsx
```diff
@@ -1,11 +1,21 @@
...
+  useIsomorphicLayoutEffect(() => {
+    fnRef.current = fn
+  }, [fn])
 
-const useSetTimeout = (fn: (...args: []) => void, delay: number) =>
   useEffect(() => {
-    const timeout = setTimeout(fn, delay)
-    return () => clearTimeout(timeout)
-  }, [fn, delay])
+    const id = setTimeout(() => fnRef.current(), ms)
```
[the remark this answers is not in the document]
minsoo-web: @manudeli 
please check this review

minsoo-web (approved):
LGTM!

configs/test-utils/src/ThrowError.tsx
manudeli: Oh that makes sense. I agree `ms` can be good devDependency to test case like this!

manudeli:
I made https://github.com/suspensive/react/issues/265 this issue will be resolved by @ssi02014 after merging this Pull Request

PR #270, opened 2023-10-28
feat(react): add experimental wrap to remove unnecessary hocs

### LINKED ISSUE #269: [Feature]: wrap(ErrorBoundary, propsWithoutChildren, (props) => {})
### Package Scope

@suspensive/react

### Description

Consider adding wrap function to wrap other component. I can remove all withXXX of suspensive apis. It can reduce number of public apis that library users have to know

### Possible Solution

```tsx
const WrappedByErrorBoundary = wrap(ErrorBoundary, PropsWithoutChildren<ComponentProps<typeof ErrorBoundary>>, (props) => {
  ...
})

```

### etc.

_No response_

manudeli:
fix #269 

# Overview

I added `wrap` as experimental feature to remove all repetition of hocs implementation and withXXX funciton's nesting.
Library size has been reduced by 12%(2,114 bytes) than before

### AS-IS dist/index.cjs
18,377 bytes

image

### TO-BE dist/index.cjs if add wrap and remove hocs
17,120 bytes

image

## I want to deprecate all hocs
In v1, I hope to add the `wrap` function as a public API and keep all hocs marked as deprecated, and in v2, I hope to remove all hocs and provide only the `wrap` function.

### Why want to deprecate all hocs
1. To reduce library size more
2. User don't have to know withSuspense, withSuspense.CSROnly, withErrorBoudnary, withAsyncBoundary, withAsyncBoundary.CSROnly, withErrorBoundaryGroup, withDelay. just have to know wrap and existing components. (number of public apis, 15(with hocs) -> 11(without hocs))
3. `wrap` function allows us to place related code in a more relevant location. We can also create a function wrapped with wrap to reuse it because calling `wrap` will make hoc.
    #### withXXX
    ```tsx
    export default withErrorBoundary(
      withSuspense(
        () => {
          return 'success'
        },
        { fallback: 'loading...' } // far position with name of withSuspense
      ),
      { fallback: () => 'error' } // far position with name of withErrorBoundary
    )
    ```
    #### wrap
    ```tsx
    export default wrap
      .ErrorBoundary({ fallback: () => 'error' })
      .Suspense({ fallback: 'loading...' })
      .on(() => {
        return 'success'
      })
    ```

## PR Checklist

5. I added documents and tests.


packages/react/src/AsyncBoundary.tsx
```diff
@@ -46,36 +47,10 @@ export const AsyncBoundary = BaseAsyncBoundary as typeof BaseAsyncBoundary & {
...
-    </AsyncBoundary>
-  )
-
-  if (process.env.NODE_ENV !== 'production') {
-    const name = Component.displayName || Component.name || 'Component'
-    Wrapped.displayName = `withAsyncBoundary(${name})`
-  }
-
-  return Wrapped
-}
```
manudeli, line 64: Unnecessary implementations of all hocs is removed by `wrap` function

websites/visualization/src/app/react/experimental/wrap/after/page.tsx
```diff
@@ -0,0 +1,23 @@
...
+        <>
+          <button onClick={() => errorBoundary.setError(new Error('trigger error by useErrorBoundary().setError'))}>
+            trigger error by useErrorBoundary().setError
+          </button>
+          <UseSuspenseQuery queryKey={['with', 1] as const} queryFn={() => api.delay(200, { percentage: 50 })} />
+        </>
+      )
+    })
+  )
+)
```
manudeli, line 23: We can compare before, after of `wrap` function 
between `websites/visualization/src/app/react/experimental/wrap/after/page.tsx` and `websites/visualization/src/app/react/experimental/wrap/before/page.tsx`

Check it please

.changeset/pretty-pumpkins-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@suspensive/react": patch
```
manudeli, line 2:
```suggestion
"@suspensive/react": minor
```

packages/react/src/index.ts
```diff
@@ -5,6 +5,8 @@ export { ErrorBoundaryGroup, withErrorBoundaryGroup, useErrorBoundaryGroup } fro
 export { AsyncBoundary, withAsyncBoundary } from './AsyncBoundary'
 export { Delay, withDelay } from './Delay'
 
+export { wrap } from './wrap'
```
minsoo-web, line 8: > I want to deprecate all hocs

I agree about this `BREAKING CHANGES`
but if this function is cover our whole hocs, what about better naming rather than `wrap`.

My suggestion is `with`

It is like jest description, using `it` 

```
it('should error when ....'
```

it is readable continually by combining `it` + `should`

### before

```tsx
withErrBoundary(ErrorBoundary, {})(MyComponent)

// wrap case
wrap(ErrorBoundary, {})(MyComponent)
```


### after

```tsx
with(ErrorBoundary, {})(MyComponent)
```

websites/visualization/src/app/react/experimental/wrap/after/page.tsx
minsoo-web: image


image


Can I compare it through the network tab? Since the user doesn't write the `wrap` function himself, do I just need to look at the performance aspect?

packages/react/src/index.ts
manudeli: I considered `with` before too. but in strict mode of JavaScript, `with` is keyword.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/with

And also, because this `wrap` function is a function of a more abstract layer, it is named `wrap`, which means it wraps more clearly than `with`. Just as `connect` in react-redux has the meaning of connecting a store and a component. `with` is impossible because it is a keyword, and is there a better naming than `wrap`?

websites/visualization/src/app/react/experimental/wrap/after/page.tsx
manudeli: > Can I compare it through the network tab? Since the user doesn't write the wrap function himself, do I just need to look at the performance aspect?
Sure, Could you please suggest how can we compare performance?

however, this change can greatly reduce the file size of the library without BREAKING CHANGE. This is a definite improvement. Additionally, the number of public APIs users have to know can be greatly reduced. Although I think this is a more important improvement than performance. So I hope that the performance comparison does not delay the merge of this feature. Before merging, it would be better to talk about the interface of `wrap (tentative name)` rather than comparing performance.

minsoo-web: Oh I see, I agree about your opinion

packages/react/src/index.ts
minsoo-web: I see, I don't know about with is keyword in strict mode. wrap is better than I thought another naming. Let's Go !

minsoo-web (approved):

manudeli:
I'm considering about wrap's another alternative interface different with this Pull Request like below too. @minsoo-web @okinawaa 

```tsx
import { ErrorBoundary, ErrorBoundaryGroup, Suspense, useErrorBoundary, wrap } from '@suspensive/react'
import { UseSuspenseQuery } from '~/components'
import { api } from '~/utils'

const logError = (error: Error) => console.error(error)

export default wrap(ErrorBoundaryGroup, { blockOutside: false })
  .wrap(ErrorBoundary, { fallback: (props) => <>{props.error.message}</>, onError: logError })
  .wrap(Suspense.CSROnly, { fallback: <>loading...</> })
  .on<{ x: string }>((props) => {

    return (
      <>
        <button onClick={() => errorBoundary.setError(new Error('trigger error by useErrorBoundary().setError'))}>
          trigger error by useErrorBoundary().setError
        </button>
        <UseSuspenseQuery queryKey={['with', 1] as const} queryFn={() => api.delay(200, { percentage: 50 })} />
      </>
    )
  })

```

minsoo-web:
> I'm considering about wrap's another alternative interface different with this Pull Request like below too. @minsoo-web @okinawaa
> 
> I think this builder pattern have good point can make users don't have to make wrap functions is nested
> 
> ```tsx
> import { ErrorBoundary, ErrorBoundaryGroup, Suspense, useErrorBoundary, wrap } from '@suspensive/react'
> import { UseSuspenseQuery } from '~/components'
> import { api } from '~/utils'
> 
> const logError = (error: Error) => console.error(error)
> 
> export default wrap(ErrorBoundaryGroup, { blockOutside: false })
>   .wrap(ErrorBoundary, { fallback: (props) => <>{props.error.message}</>, onError: logError })
>   .wrap(Suspense.CSROnly, { fallback: <>loading...</> })
>   .on<{ x: string }>((props) => {
>     const errorBoundary = useErrorBoundary()
> 
>     return (
>       <>
>         <button onClick={() => errorBoundary.setError(new Error('trigger error by useErrorBoundary().setError'))}>
>           trigger error by useErrorBoundary().setError
>         </button>
>         <UseSuspenseQuery queryKey={['with', 1] as const} queryFn={() => api.delay(200, { percentage: 50 })} />
>       </>
>     )
>   })
> ```
> 
> then this is way also okay
> 
> ```tsx
> const wrapErrorBoundary = wrap(ErrorBoundary, { fallback: (props) => <>{props.error.message}</>, onError: logError })
> const Page = wrapErrorBoundary.on((props) => {
>     const errorBoundary = useErrorBoundary()
> 
>     return (
>       <>
>         <button onClick={() => errorBoundary.setError(new Error('trigger error by useErrorBoundary().setError'))}>
>           trigger error by useErrorBoundary().setError
>         </button>
>         <UseSuspenseQuery queryKey={['with', 1] as const} queryFn={() => api.delay(200, { percentage: 50 })} />
>       </>
>     )
>   })
> ```

**I think it's a really good dx**. In fact, it's a functional type, so I think giving a pipe-like feeling can be a better experience for the developer.

On the other hand, I'm worried that it might give me a `callback hell` experience like the case where promise was used consecutively, such as then.then.then. 

The role is to remove unnecessarily declared `wrappers`, but is it rather lowering the readability too much? 
**That's what I'm worried about**

To sum up, I think it's really good to provide a builder pattern with chaining, but I think it might rather hurt the dx if it's used in a row.

manudeli:
@minsoo-web Personally, I sometimes prefer the then keyword over the async/await keyword because it allows chaining and makes it easier to read the promises in order. So, if I need to use multiple suspensive components such as ErrorBoundary and Suspense together, I think I will use chaining rather than nesting the wrap function. Nevertheless, if you provide chaining of the wrap function like this, library users can choose whether to use chaining. So I want to provide this wrap chaining.

We can provide below chaining of wrap

```tsx
import { ErrorBoundary, ErrorBoundaryGroup, Suspense, useErrorBoundary, wrap } from '@suspensive/react'
import { UseSuspenseQuery } from '~/components'
import { api } from '~/utils'

const logError = (error: Error) => console.error(error)

export default wrap(ErrorBoundaryGroup, { blockOutside: false })
  .wrap(ErrorBoundary, { fallback: (props) => <>{props.error.message}</>, onError: logError })
  .wrap(Suspense.CSROnly, { fallback: <>loading...</> })
  .on<{ x: string }>((props) => {

    return (
      <>
        <button onClick={() => errorBoundary.setError(new Error('trigger error by useErrorBoundary().setError'))}>
          trigger error by useErrorBoundary().setError
        </button>
        <UseSuspenseQuery queryKey={['with', 1] as const} queryFn={() => api.delay(200, { percentage: 50 })} />
      </>
    )
  })
```

or also below is okay without chaining

```tsx
import { ErrorBoundary, ErrorBoundaryGroup, Suspense, useErrorBoundary, wrap } from '@suspensive/react'
import { UseSuspenseQuery } from '~/components'
import { api } from '~/utils'

const logError = (error: Error) => console.error(error)

export default wrap(ErrorBoundaryGroup, { blockOutside: false }).on(
  wrap(ErrorBoundary, { fallback: (props) => <>{props.error.message}</>, onError: logError }).on(
    wrap(Suspense.CSROnly, { fallback: <>loading...</> }).on<{ x: string }>((props) => {

      return (
        <>
          <button onClick={() => errorBoundary.setError(new Error('trigger error by useErrorBoundary().setError'))}>
            trigger error by useErrorBoundary().setError
          </button>
          <UseSuspenseQuery queryKey={['with', 1] as const} queryFn={() => api.delay(200, { percentage: 50 })} />
        </>
      )
    })
  )
)

```

I prefer chaining but library users can choose 2 ways both.

manudeli:
I change this Pull Request as draft. because I made this https://github.com/suspensive/react/pull/280 that update wrap as builder pattern to remove nested wrap functions with combination importing all public apis

packages/react/src/wrap.ts
```diff
@@ -0,0 +1,110 @@
...
+  | WrapperItem<typeof Suspense.CSROnly>
+  | WrapperItem<typeof ErrorBoundary>
+  | WrapperItem<typeof ErrorBoundaryGroup>
+  | WrapperItem<typeof AsyncBoundary>
+  | WrapperItem<typeof AsyncBoundary.CSROnly>
+  | WrapperItem<typeof Delay>
+
+class WrapWithoutCSROnly {
+  constructor(private wrappers: Wrapper[]) {}
+  Suspense = (props: PropsWithoutChildren<ComponentProps<typeof Suspense>> = {}) => {
```
minsoo-web, line 28: Do we need to provide default parameter??

minsoo-web (commented):
I think you used default parameters often, is there a reason?

packages/react/src/wrap.ts
manudeli: I thought wrap.Suspense, wrap.Suspense.CSROnly, wrap.Delay don't need props object

```tsx
const Example = wrap.Suspense(/* no parameter is available*/).on(() => {
  return 'success'
})
```

minsoo-web: then how about optional type define?

```suggestion
  Suspense = (props?: PropsWithoutChildren<ComponentProps<typeof Suspense>>) => {
```

image

It can prevent unnecessary variable allocation.

manudeli: Not all wrap.XXX have optional props object. optional props object is only for wrap.Suspense, wrap.Suspense.CSROnly, wrap.Delay. wrap.ErrorBoundary, wrap.ErrorBoundaryGroup require props object

manudeli:
@okinawaa It's experimental feature. I merge this to make new release contain this feature quickly. this will make us get feedback to improve this experimental feature easily and we can fix it later.

PR #282, opened 2023-10-31
chore: Pin node version with volta

### LINKED ISSUE #171: [Feature]: How about to use volta?
### **Package Scope**

- [ ] New package

Package name:

### **Overview**

- How about using volta instead of nvm?

sungh0lim:
fix #171 

# Overview

I pinned the node version using [Volta](https://docs.volta.sh/guide/getting-started).

## PR Checklist


minsoo-web (commented):
We have specified how to set node version in [this document](https://github.com/suspensive/react/blob/main/CONTRIBUTING.md#prerequisites). 

Adding ways to support `Volta` might also be a good way, but it is a pessimistic position about whether it is necessary to provide a guide to new tools from the perspective of contributors, not users.

@manudeli 
What's your opinion?

manudeli (approved):
@sungh0lim @minsoo-web 
I am considering the following:

1. A way to avoid having to install volta and configure it separately not like nvm.
2. A way to use it regardless of both volta and nvm users.
3. I looked to see if volta could be used as a devDependency, but it is not possible.
4. Volta does not yet support pnpm in a stable state.

Despite these concerns, specifying volta in the monorepository's package.json does not seem to be a problem for both nvm and volta users, so I want to merge this Pull Request first.  

@sungh0lim Please create another new Pull Request to add more detailed explanations to CONTRIBUTION.md for volta users.

sungh0lim:
@manudeli, @minsoo-web  Thank you for comment.

It is unfortunate that volta does not support pnpm. However, I think users using Volta can change the node version more quickly through this PR.

Like above comments, we need an explanation of volta in CONTRIBUTING.md. I'll try to add it.
Can I merge this PR? If so, I will merge it.

PR #325, opened 2023-11-09
perf(react): remove useKey to reduce ErrorBoundaryGroup's size

tooooo1:
# Overview

~- Renames `useKey` to `useRefreshKey` for clarity.~
~- **Clarifies purpose:** useRefreshKey indicates it's for triggering re-renders, not for key management.~
~- **Aids understanding:** New name instantly informs about the hook's function without confusion.~
~- **Promotes clean code:** More descriptive name aligns with clean coding practices.~
~- **Enhances maintainability:** Clear names make future changes and collaboration easier.~

~or `useCounter`, `useForceUpdate`,etc...~

- Removed `useKey` custom hook.
- Implemented `useReducer` hook with `increase` reducer for managing the reset key state.

## PR Checklist


packages/react/src/hooks/useRefreshKey.ts
```diff
@@ -1,6 +1,6 @@
 import { useCallback, useState } from 'react'
 
-export const useKey = (): [number, () => void] => {
+export const useRefreshKey = (): [number, () => void] => {
   const [key, setKey] = useState(0)
   const refresh = useCallback(() => setKey((prev) => prev + 1), [])
```
manudeli, line 5: refresh -> count

manudeli (commented):
@okinawaa @minsoo-web In my opinion, useCount is proper

then, rename it as resetKey in file using useCount

```jsx
const [resetKey, resetKeyUp] = useCount()
```

minsoo-web:
I think use counter have to can have minus caes this hook only cover plus count case

manudeli:
> I think useCounter have to can have minus caes this hook only cover plus count case

Then useIncreasingNumber?

```tsx
const increase = (prev: number) => prev + 1
export const useIncreasingNumber = () => {
  const [value, setValue] = useState(0)

  return useMemo(
    () => ({
      value,
      increase: () => setValue(increase),
    }),
    [value]
  )
}
```

tooooo1:
I don't think there's any reason to change the name. I tried using the reducer right away.

```tsx
const [resetKey, reset] = useReducer(increase, 0)
```

packages/react/src/hooks/useKey.ts
```diff
@@ -1,8 +0,0 @@
-import { useCallback, useState } from 'react'
-
-export const useKey = (): [number, () => void] => {
-  const [key, setKey] = useState(0)
-  const refresh = useCallback(() => setKey((prev) => prev + 1), [])
```
manudeli, line 5: Now we don't memo refresh function

manudeli (approved):
@tooooo1 Cool, We can get
1. no memoization refresh function
2. no useKey don't be used twice.
And I think your code is more intuitive than before. thanks! 👍

packages/react/src/ErrorBoundaryGroup.tsx
```diff
@@ -18,22 +18,24 @@ export type ErrorBoundaryGroupProps = PropsWithChildren<{
...
 
+const increase = (prev: number) => prev + 1
+
 /**
  * ErrorBoundaryGroup is Component to manage multiple ErrorBoundaries
  * @see {@link https://suspensive.org/docs/react/ErrorBoundaryGroup}
  */
 export const ErrorBoundaryGroup = ({ blockOutside = false, children }: ErrorBoundaryGroupProps) => {
-  const [resetKey, reset] = useKey()
+  const [resetKey, reset] = useReducer(increase, 0)
```
manudeli, line 28: Memory will be better because we just use `increase` function every time.
`useReducer with `increase` is more intuitive than `useKey`. 👍

manudeli:
@minsoo-web I think this is definitely better than before for intuitiveness and performance both.
Could you merge this Pull Request if you agree it?

packages/react/src/ErrorBoundaryGroup.tsx
minsoo-web: COOL!! We didn't have to make our own new hook! That's so cool

minsoo-web (approved):
Many Thanks !!

Issue #295, opened 2023-11-04
[Feature]: deprecate AsyncBoundary in v1, remove AsyncBoundary in v2


manudeli:
### Package Scope

@suspensive/react

### Description

I think we need to remove AsyncBoundary in @suspensive/react v2.
1. current ErrorBoundary, ErrorBoundaryGroup, ErrorBoundaryFallbackProps, useErrorBoundary, useErrorBoundaryFallbackProps affect AsyncBoundary. it makes confusion to understand because of naming mismatch
    - useErrorBoundary, ErrorBoundaryFallbackProps, useErrorBoundaryFallbackProps affect to AsyncBoundary but its name is different
        ```tsx
        const Example = withAsyncBoundary(
          () => {
            // we can use useErrorBoundary in AsyncBoundary. it can't be useAsyncBoundary
            const errorBoundary = useErrorBoundary()
        
            errorBoundary.setError
        
            return <>success</>
          },
          {
            pendingFallback: 'loading...'
            // we can use ErrorBoundaryFallbackProps for AsyncBoundary's rejectedFallback.props . it can't be AsyncBoundaryRejectedFallbackProps
            rejectedFallback: (props: ErrorBoundaryFallbackProps) => {
              // we can use useErrorBoundaryFallbackProps in AsyncBoundary's rejectedFallback. it can't be useAsyncBoundaryRejectedFallbackProps
              const errorBoundaryFallbackProps = useErrorBoundaryFallbackProps()
        
              errorBoundaryFallbackProps.error
              errorBoundaryFallbackProps.reset
        
              return 'error'
            },
          }
        )
        ```
    - ErrorBoundaryGroup affect AsyncBoundary
        ```tsx
        const Example () => (
          <ErrorBoundaryGroup>
            <ErrorBoundary />
            <AsyncBoundary /> // it will be affected by ErrorBoundaryGroup even though this is just `"ErrorBoundary"Group` not `"AsyncBoundary"Group`
          </ErrorBoundaryGroup>
        )
        ```
2. because of AsyncBoundary's new interface like pendingFallback, rejectedFallback, we need to know more number of interfaces.
[...]

PR #345, opened 2023-11-11
fix(react): deprecate AsyncBoundary

manudeli:
related #295

# Overview

I deprecate AsyncBoundary to remove it in v2. because of this https://github.com/suspensive/react/issues/295

When I first created AsyncBoundary, I thought it was a good implementation because it allowed me to use Suspense and ErrorBoundary at once, but due to the problems raised by this issue, I thought it was right to deprecate it.

However, in order to use useErrorBoundary, useErrorBoundaryFallbackProps, ErrorBoundaryFallbackProps, ErrorBoundaryGroup, and useErrorBoundaryGroup without confusion, I thought it would be right to leave the AsyncBoundary implementation to the library users themselves.

## PR Checklist


minsoo-web (commented):
Rather than deleting it at once, how about using JSDoc to inform you that it is going to be deprecated ? I think it would be a bit weird to say BREAKING_CHANGE if you suddenly import a function you used well

manudeli:
@minsoo-web I added deprecated jsdoc to promote library users not to use AsyncBoundary https://github.com/suspensive/react/pull/345/files#diff-ec215d2d4200c2150aa113534215f15d2b1127b486f356d3082551ff92fcd48cR39

You mean this right?

manudeli:
@minsoo-web I left comment to respond your question, I want to get another review or merge this Pull Request. could you please re-review?

docs/src/pages/docs/react/AsyncBoundary.en.mdx
minsoo-web, line 1: My intention was to distribute only the version that tells you in advance that `AsyncBoundary` **will be** gone, and then I thought of making it impossible to import itself in the future. 
It is impossible to import it at all if they do the version up now, but I'm afraid it's too big change for the user who was using it well.

minsoo-web: I'm sorry for the late response 😢.

manudeli: In the next version (1.***) of @suspensive/react, AsyncBoundary will not disappear and library users will be able to use it. It's just deprecated as jsdoc. I need example or code or etc. a more specific suggestion.

minsoo-web (approved):
Lets do this 🚀 🚀

PR #367, opened 2023-11-15
feat(await): rename @suspensive/react-await to @suspensive/await

manudeli:
# Overview

## PR Checklist


packages/await/package.json
minsoo-web, line 1: This package is dependent on react, is there a reason for excluding the `react-` prefix?

manudeli: I thought name 'suspensive' already have meaning depending on 'react'. so I just want to remove name 'react' from naming '@suspensive/react-await'. but I am also confused that name '@supensive/react' have name 'react' yet. so I can't be confident on naming of '@suspensive/await'

manudeli:
@minsoo-web I've been thinking about this topic for a long time. I decided to use @suspensive/react-*. We have to be careful when using @suspensive/* because it is an irreversible decision. So, like this, I have decided. I close this Pull Request

PR #462, opened 2023-12-12
ci(.github/pnpm/setup-node.yml): pnpm action

manudeli:
# Overview

add pnpm action to keep consistency

## PR Checklist


minsoo-web (commented):
It's such a great job!! 

It was really impressive that we didn't have to use the composite to repeat it. 
It's an honor for me to learn from this opportunity, too. 

However, we didn't deploy this flow to market, we're only going to use it locally, 
so I think there will be **more actions** that will be repeated in the future. 

Why don't we improve the folder naming a little bit?

`.github/pnpm/setup-node/action.yml`

// after
`.github/workflows/actions/pnpm-setup-node.yml`

or

`.github/actions/pnpm-setup-node.yml`

---

To make the point, I think that file should be named a role rather than `action.yml`, and the top folder should be `actions`

.github/workflows/ci.yml
```diff
@@ -22,14 +22,7 @@ jobs:
...
       - uses: actions/checkout@v4
-      - uses: pnpm/action-setup@v2
-        with:
-          version: 8.12.0
-      - uses: actions/setup-node@v4
-        with:
-          cache: 'pnpm'
-          cache-dependency-path: 'pnpm-lock.yaml'
-          node-version-file: '.nvmrc'
+      - uses: ./.github/pnpm/setup-node
```
manudeli, line 25: @minsoo-web why is your suggestion better than this? I need more reason to change it

minsoo-web (approved):
Thank you for accepting my opinion!!

PR #463, opened 2023-12-12
fix(tsup): add splitting to reduce bundle size dramatically

manudeli:
# Overview

add splitting to reduce bundle size dramatically for esm

#### AS-IS (721KB)
Screenshot 2023-12-12 at 8 52 26 PM

#### TO-BE (407KB)
Screenshot 2023-12-12 at 8 53 09 PM

## PR Checklist

### DIFF
--- a/.changeset/smooth-birds-sip.md
+++ b/.changeset/smooth-birds-sip.md
@@ -1,5 +1,8 @@
 ---
-"@suspensive/tsup": patch
+"@suspensive/react": patch
+"@suspensive/react-query": patch
+"@suspensive/react-await": patch
+"@suspensive/react-image": patch
 ---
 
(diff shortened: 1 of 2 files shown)


minsoo-web (approved):
Seeing that the bundle size has been dramatically reduced, I wanted to apply it right away 🚀🚀. 

Since we're working on reducing the **bundle size**, why don't we set the `treeshake` option and the `minify` and `minifyWhitespace` options? 

I know they're all options that contribute to optimizing the bundle size

for example treeshake option (Delay.js)

image
> left: turn on `treeshake`

---

default (on this PR) (--force option build)
image

apply (treeshake, minify, minifyWhitespace)
image

Issue #446, opened 2023-12-08
[Feature]: add DevMode


manudeli:
### Package Scope

@suspensive/react, etc

### Description

To develop fallback of Suspense or ErrorBoundary easily, we need to make suspending or throw error on purpose

### Possible Solution

```tsx
const DevOnly = {
  Suspend: () => {
    if (process.env.NODE_ENV !== 'production') {
      throw new Promise(() => {})
    }
    return <></>
  },
  Throw: ({ error, afterMs = 0 }: { error: Error, afterMs: number }) => {
    const [isNeedError, setIsNeedError] = useState(after === 0 ? true : throwErrorIsNeed.current)
    if (process.env.NODE_ENV !== 'production' && isNeedError) {
      throw error
    }
    useTimeout(() => setIsNeedError(true), afterMs)
    return <></>
  },
}
```

PR #470, opened 2023-12-13
feat(react): add DevMode

manudeli:
fix: #446

# Overview

@okinawaa @minsoo-web I want to add new feature `DevMode` in @suspensive/react or in new package @suspensive/react-devtools

Before add new feature, I want discuss this without explaining it. because I want to make DevMode's user understand it easily without studying

```tsx
'use client'

import type { ErrorBoundaryFallbackProps } from '@suspensive/react'
import { DevMode, ErrorBoundary, Suspense, devMode } from '@suspensive/react'
import { Area, Box, Button, Spinner } from '~/components/uis'

export default function Page() {
  return (
    <Area title="devMode">
      <Area title="<Suspense devMode={{ showFallback: true }} />">
        <Suspense fallback={<Spinner />} devMode={{ showFallback: true }}>
          <Box.Success>children</Box.Success>
        </Suspense>
      </Area>

      <Area title="<ErrorBoundary devMode={{ showFallback: true }} />">
        <ErrorBoundary fallback={ErrorBoundaryFallback} devMode={{ showFallback: true }}>
          <Box.Default>children</Box.Default>
        </ErrorBoundary>
      </Area>
      <Area title="<ErrorBoundary devMode={{ showFallback: { after: 2000 } }} />">
        <ErrorBoundary fallback={ErrorBoundaryFallback} devMode={{ showFallback: { after: 2000 } }}>
          <Box.Default>children</Box.Default>
        </ErrorBoundary>
      </Area>
    </Area>
  )
}

const ErrorBoundaryFallback = ({ error, reset }: ErrorBoundaryFallbackProps) => (
  <Box.Error>
    {error.message}
    <Button onClick={reset}>↻</Button>
  </Box.Error>
)

```

## PR Checklist


minsoo-web:
First of all, I'm sorry for the late confirmation.

First of all, I understood by looking **at the code**, but from the user's point of view, there seems to be something that needs to **_be improved by showing documents and example cases_**. 

From my understanding, `DevMode`'s role seems to be a tool to help you control 

> "when you want to see the fallback of `Suspense`, `Errorboundry` **on purpose**," 

is that right?

I think `DevMode`'s naming made it harder to understand for now. 
I think `Redux` or `react-query` are examples of that

And assuming that we usually use developer tools, I think we **only declared it once** at the top and tested the development environment using only that tool. 
But now that I control each by calling `DevMode` components **to a specific `Suspense` and `Errorboundry`**, I thought it might be a little inconvenient for users. 
Because, I think I'll **have to intentionally erase the code** when I go to production.

First of all, I think that if I understand what I meant, it would give me a **great DX**. 
But I thought that I should give you a guide through **clear examples** and **documents** to improve.

minsoo-web:
And visualization app is also production, user can't see Devmode 😢

manudeli:
Note to remember idea: how about make each Suspense, ErrorBoundary contain DevMode

manudeli:
@minsoo-web I reflect our discussion in #481. check it out please.
I want to get confirmation of this new public apis before writing official docs

packages/react/src/DevMode.tsx
```diff
@@ -0,0 +1,110 @@
...
+> = TComponentProps & {
+  /**
+   * @experimental This is experimental feature.
+   */
+  devMode?: TDevModeProps
+}
+
+type Sync = () => void
+class SuspensiveDevMode {
+  constructor(public is = false) {}
```
manudeli, line 19: SuspensiveDevMode.is will be always false before developer's clicking `<DevMode/>`.
This will prevent misunderstanding if SuspensiveDevMode on or off now.

minsoo-web (approved):
It's a good implementation of what we discussed! I think it's so cool 🚀 . 

Rather than talking about the current useCase, I'd like to try and improve this feature, even if it's experimental.

manudeli:
[Feature] svg logo for SuspensiveDevTools need

PR #535, opened 2024-01-01
feat(react): add svg logo for `<DevMode/>`

### LINKED ISSUE #503: [Feature] svg logo for DevMode need
[Feature] svg logo for SuspensiveDevTools need

_Originally posted by @manudeli in https://github.com/suspensive/react/issues/470#issuecomment-1868241138_

manudeli:
fix #503 

# Overview

I add svg logo for `<DevMode/>`

chrome-capture-2024-0-2

## PR Checklist


packages/react/src/DevMode.tsx
```diff
@@ -26,31 +26,415 @@ export const DevMode = ({ position = 'bottomRight' }: DevModeProps) => {
...
   const devMode = useDevModeObserve()
-  if (devMode == null) {
-    return null
+  if (process.env.NODE_ENV !== 'production' && devMode != null) {
+    return (
+      <svg
+        role={`Suspensive.DevMode-${devMode.is ? 'on' : 'off'}`}
+        onClick={devMode.is ? devMode.off : devMode.on}
+        onMouseEnter={() => setIsHover(true)}
+        onMouseLeave={() => setIsHover(false)}
```
minsoo-web, line 35: I'm wondering if we need to call `useState` to control hover state.

Since we can't control the hover state from **inline CSS**, I guess it's true that we can only give the hover effect by controlling the state via `useState` at this point.

But in essence, I don't see **how adding a hover state to this UI can provide any value to the user**. 

Personally, I think a hover effect is meant **to encourage clicks**, but since `DevMode` is something that the user **will do on their own when they need to**

Personally, I think the hover effect is used to encourage clicking, but 'DevMode' assumes that the user will press it when they need to.

To summarize, **do we need to use `useState` to control the hover state?** is my opinion.

minsoo-web (commented):
우리 로고가 들어가서 너무 기뻐!! 하지만, 우리 로고가 흰색 베이스라.. 배경색이 필요할 것 같은 느낌이 들어. 흰 화면에서 테스트해본 영상을 첨부할게

https://github.com/suspensive/react/assets/57122180/ab2f503c-5bb1-4ee8-b2e1-9a11c5180464

packages/react/src/DevMode.tsx
```diff
@@ -26,31 +26,415 @@ export const DevMode = ({ position = 'bottomRight' }: DevModeProps) => {
...
+          <linearGradient
+            id="paint0_linear_997_2"
+            x1="543.869"
+            y1="85.5625"
+            x2="543.869"
+            y2="275.816"
+            gradientUnits="userSpaceOnUse"
+          >
+            <stop stopColor="#FF0000" />
+            <stop offset="1" stopColor="#FF0000" stopOpacity="0" />
```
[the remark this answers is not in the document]
manudeli: It is for gradient layer-blur. You can see it in our figma
image

I just use that red color to differ other color definitely

packages/react/src/DevMode.tsx
manudeli: I found there is no svg or no useState code in built bundle. you can check it in websites/visualization with DevMode.

image

Then isn't it better to support better UX in development mode?

packages/react/src/DevMode.tsx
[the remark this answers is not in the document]
minsoo-web: Thanks for the confirmation!!!

packages/react/src/DevMode.tsx
minsoo-web: Got it! Thanks for the confirmation!

minsoo-web (commented):
How can we improve SVGs not showing up on a screen with a white background? 
How about a black background color?

manudeli:
> How can we improve SVGs not showing up on a screen with a white background? How about a black background color?

I added backgroundColor: black for `<DevMode/>` in [5f03a08](https://github.com/suspensive/react/pull/535/commits/5f03a08429c8e808ca7f316249b08ec9f7e9ac6f)
Thanks for your eagle eyes

minsoo-web (approved):
LGTM!! Lets Go !!

PR #539, opened 2024-01-02
ci(eslint): add plugin:@typescript-eslint/strict-type-checked

manudeli:
# Overview

I added @typescript-eslint/strict-type-checked to lint more strictly

https://x.com/TkDodo/status/1741443033642487897?s=20

## PR Checklist

### DIFF
--- a/configs/tsup/tsup.config.ts
+++ b/configs/tsup/tsup.config.ts
@@ -1,4 +1,4 @@
 import { defineConfig } from 'tsup'
-import { options } from './src'
+import { options } from './src/index.js'
 
 export default defineConfig(options)
--- a/packages/react-query/src/useSuspenseInfiniteQuery.ts
+++ b/packages/react-query/src/useSuspenseInfiniteQuery.ts
@@ -141,14 +141,6 @@ export function useSuspenseInfiniteQuery<
   TData = TQueryFnData,
   TQueryKey extends QueryKey = QueryKey,
 >(
-  options: Omit<UseSuspenseInfiniteQueryOptions<TQueryFnData, TError, TData, TQueryKey>, 'enabled'>
-): UseSuspenseInfiniteQueryResultOnSuccess<TData>
-export function useSuspenseInfiniteQuery<
-  TQueryFnData = unknown,
-  TError = unknown,
-  TData = TQueryFnData,
-  TQueryKey extends QueryKey = QueryKey,
->(
   options: Omit<UseSuspenseInfiniteQueryOptions<TQueryFnData, TError, TData, TQueryKey>, 'enabled'> & {
     enabled?: true
   }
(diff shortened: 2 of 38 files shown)


configs/tsup/tsup.config.ts
minsoo-web, line 1: This could be a separate issue, but do we really need to make tsup a ts? JSDoc's `@type` allows us to reason about it, and I don't think it would be inconvenient, but would it create a build dependency? I was thinking

packages/react-query/src/useSuspenseInfiniteQuery.ts
```diff
@@ -135,14 +135,6 @@ export function useSuspenseInfiniteQuery<
 ): UseSuspenseInfiniteQueryResultOnSuccess<TData> | UseSuspenseInfiniteQueryResultOnLoading
 
 // arg1: options
```
minsoo-web, line 137: Is this a necessary comment?

packages/react/src/Suspense.spec.tsx
```diff
@@ -5,17 +5,17 @@ import { createElement } from 'react'
 import { beforeEach, describe, expect, it } from 'vitest'
 import { Suspense, withSuspense } from '.'
 
-const resetBeforeEach = () => beforeEach(Suspend.reset)
+const resetBeforeEach = () => beforeEach(() => Suspend.reset())
```
minsoo-web, line 8: I'm wondering what you're getting by modifying it, is it unnecessary anonymous function declarations and calls?

minsoo-web (commented):
I left a few comments! That's a Lint setting I didn't know about, thanks for letting me know!!!

packages/react-query/src/useSuspenseInfiniteQuery.ts
manudeli: Only for develop library, not user. This comment is only for me or other maintainer to differ 3 sections of overrode function types(arg1, arg2?, arg3?), but this comment won't be contained

configs/tsup/tsup.config.ts
manudeli: You are right, but I thought if @suspensive/tsup need to be extended, TypeScript will be helpful for it.
It's not problem if you change it as JavaScript with `@types`.

packages/react/src/Suspense.spec.tsx
manudeli: https://typescript-eslint.io/rules/unbound-method/

I don't think it's really problem but, I just want to avoid their eslint error if there is chance to make error by it.

image

configs/tsup/tsup.config.ts
minsoo-web: I understand, I just thought the tsup.config file was simply for the config file, so that was my comment, thanks!!

packages/react/src/Suspense.spec.tsx
minsoo-web: Thanks for the explanation, I didn't realize that was an eslint setting! Thanks for the confirmation!

minsoo-web (approved):
LGTM!! Many Thanks!!

PR #555, opened 2024-01-04
feat(react): add fallback prop of Delay

### LINKED ISSUE #480: [Feature]: add fallback of Delay
### Package Scope

@suspensive/react

### Description

Add fallback of Delay

### Possible Solution

_No response_

### etc.

_No response_

manudeli:
fix #480 

# Overview

I add fallback prop of Delay

## PR Checklist


packages/react/src/Delay.tsx
```diff
@@ -1,17 +1,21 @@
...
+   */
+  fallback?: ReactNode
 }
 
-export const Delay = ({ ms, children }: DelayProps) => {
+export const Delay = ({ ms, children, fallback }: DelayProps) => {
   const defaultProps = useContext(DelayDefaultPropsContext)
   const delayMs = ms ?? defaultProps.ms ?? 0
   const [isDelayed, setIsDelayed] = useState(delayMs === 0)
   useTimeout(() => setIsDelayed(true), delayMs)
```
minsoo-web, line 17: The intuitive nuance of isDelayed is that it has a stronger sense of "was it delayed?" (past tense) and the actual usage is stronger "did you wait for all the delayed time?", so why not change it to isDelaying and swap the true false?

> 한국말로, 지연되었는지?

```suggestion
  const [isDelaying, setIsDelaying] = useState(delayMs !== 0)
  useTimeout(() => setIsDelaying(false), delayMs)
```

It may not make a huge difference in readability, but let me know what you think!

packages/react/src/Delay.tsx
```diff
@@ -1,17 +1,21 @@
...
-  return <>{isDelayed ? children : null}</>
+  return <>{isDelayed ? children : typeof fallback === 'undefined' ? defaultProps.fallback : fallback}</>
```
minsoo-web, line 18: Rather than using nested trinomials, why not just define and use the fallback once above, or even use an early return strategy to simplify the return statement!

packages/react/src/Delay.tsx
```diff
@@ -1,17 +1,21 @@
...
+   * @experimental This is experimental feature.
+   */
+  fallback?: ReactNode
 }
 
-export const Delay = ({ ms, children }: DelayProps) => {
+export const Delay = ({ ms, children, fallback }: DelayProps) => {
   const defaultProps = useContext(DelayDefaultPropsContext)
   const delayMs = ms ?? defaultProps.ms ?? 0
   const [isDelayed, setIsDelayed] = useState(delayMs === 0)
```
minsoo-web, line 16: Shouldn't we think about using the `Delay` component in the Server Component? 
Of course, our `tsup` config file seems to have 'use client' attached to everything, but I just wanted to ask just to be sure!

minsoo-web (commented):
I left a few questions for you, check them out!

packages/react/src/Delay.tsx
manudeli: Cool, I renamed it as isDelaying in [aafcfc6](https://github.com/suspensive/react/pull/555/commits/aafcfc6f0f560647a18918b0dc74d8068bece787)

packages/react/src/Delay.tsx
```diff
@@ -1,17 +1,27 @@
...
+export const Delay = ({ ms, children, fallback }: DelayProps) => {
   const defaultProps = useContext(DelayDefaultPropsContext)
   const delayMs = ms ?? defaultProps.ms ?? 0
-  const [isDelayed, setIsDelayed] = useState(delayMs === 0)
-  useTimeout(() => setIsDelayed(true), delayMs)
-  return <>{isDelayed ? children : null}</>
+
+  if (process.env.NODE_ENV !== 'production') {
+    assert(delayMs >= 0, 'Delay: ms prop should be greater than or equal to 0')
+  }
```
manudeli, line 20: @minsoo-web I added assert for `ms` prop in development mode.

packages/react/src/Delay.tsx
```diff
@@ -1,17 +1,27 @@
...
+
+  const [isDelaying, setIsDelaying] = useState(delayMs > 0)
+  useTimeout(() => setIsDelaying(false), delayMs)
+  return isDelaying ? (typeof fallback === 'undefined' ? defaultProps.fallback : fallback) : children
```
manudeli, line 24: I removed unnecessary jsx `<></>`

> Rather than using nested trinomials, why not just define and use the fallback once above, or even use an early return strategy to simplify the return statement!

@minsoo-web isn't it simpler than before? I don't want to make unnecessary variable name if possible.

packages/react/src/Delay.tsx
manudeli: Could you more explain?

packages/react/src/Delay.tsx
```diff
@@ -1,17 +1,30 @@
...
+  if (process.env.NODE_ENV !== 'production') {
+    assert(ms >= 0, 'Delay: ms prop should be greater than or equal to 0')
+  }
+
+  const [isDelaying, setIsDelaying] = useState(ms > 0)
+  useTimeout(() => setIsDelaying(false), ms)
+
+  const fallback = typeof props.fallback === 'undefined' ? defaultProps.fallback : props.fallback
+
+  return isDelaying ? fallback : props.children
```
manudeli, line 27: I removed unnecessary jsx `<></>`

> Rather than using nested trinomials, why not just define and use the fallback once above, or even use an early return strategy to simplify the return statement!

@minsoo-web I made it as like this

packages/react/src/Delay.tsx
```diff
@@ -1,17 +1,30 @@
...
   const defaultProps = useContext(DelayDefaultPropsContext)
-  const delayMs = ms ?? defaultProps.ms ?? 0
-  const [isDelayed, setIsDelayed] = useState(delayMs === 0)
-  useTimeout(() => setIsDelayed(true), delayMs)
-  return <>{isDelayed ? children : null}</>
+
+  const ms = props.ms ?? defaultProps.ms ?? 0
+  if (process.env.NODE_ENV !== 'production') {
+    assert(ms >= 0, 'Delay: ms prop should be greater than or equal to 0')
+  }
```
manudeli, line 20: @minsoo-web I added assert for `ms` prop in development mode.

packages/react/src/Delay.tsx
```diff
@@ -1,17 +1,33 @@
...
-  useTimeout(() => setIsDelayed(true), delayMs)
-  return <>{isDelayed ? children : null}</>
+
+  const ms = props.ms ?? defaultProps.ms ?? 0
+  if (process.env.NODE_ENV !== 'production') {
+    assert(
+      [props.ms, defaultProps.ms].every((item = 0) => item >= 0),
+      'Delay: ms prop should be greater than or equal to 0'
+    )
+  }
```
manudeli, line 23: @minsoo-web I added assertion for `ms` prop in development mode.

packages/react/src/Delay.tsx
minsoo-web: You shouldn't even bother with this review anymore! 
I was wrong to leave it.

packages/react/src/Delay.tsx
minsoo-web:
```suggestion
  if (process.env.NODE_ENV !== 'production') {
    assert(ms > 0, 'Delay: ms prop should be greater than or equal to 0')
  }
```

이렇게 해도 되지 않아?? 바로 윗 줄에서 `ms`를 `nullish coalescing operator` (??) 을 통해 제어했으니깐 똑같은 처리라고 생각해

packages/react/src/Suspensive.tsx
```diff
@@ -14,6 +15,14 @@ export class Suspensive {
...
 
   constructor(config: { defaultOptions?: Suspensive['defaultOptions'] } = {}) {
+    if (process.env.NODE_ENV !== 'production') {
+      if (typeof config.defaultOptions?.delay?.ms === 'number') {
+        assert(
+          config.defaultOptions.delay.ms > 0,
+          'Suspensive: config.defaultOptions.delay.ms should be greater than 0'
+        )
+      }
+    }
```
manudeli, line 25: But I thought defaultOptions.delay.ms should be > 0. so I made this assertions differently with ms prop of Delay

```suggestion
    if (process.env.NODE_ENV !== 'production') {
      if (typeof config.defaultOptions?.delay?.ms === 'number') {
        assert(
          config.defaultOptions.delay.ms > 0,
          'Suspensive: config.defaultOptions.delay.ms should be greater than 0'
        )
      }
    }
```

packages/react/src/Delay.tsx
```diff
@@ -1,17 +1,30 @@
...
+  fallback?: ReactNode
 }
 
-export const Delay = ({ ms, children }: DelayProps) => {
+export const Delay = (props: DelayProps) => {
+  if (process.env.NODE_ENV !== 'production') {
+    if (typeof props.ms === 'number') {
+      assert(props.ms > 0, 'Delay: ms prop should be greater than or equal to 0')
+    }
+  }
```
manudeli, line 19: And I assert that ms props of Delay should be greater than or equal to 0. differently with Suspensive defaultOptions

manudeli (commented):
@minsoo-web I left my intentions

packages/react/src/Delay.tsx
manudeli: I made it as variable in assert.ts

packages/react/src/Suspensive.tsx
manudeli: I made it as variable in assert.ts

minsoo-web (approved):
Thank you for reflecting my review!! I also checked your comments!! Lets Go!! 🚀🚀

Issue #583, opened 2024-01-09
[Bug]: `forwardRef is not defined` client-side error in Next.js production builds


relsunkaev:
### Package Scope

@suspensive/react

### Bug description

After updating to 1.20.10 the app throws a client-side error

```
ReferenceError: forwardRef is not defined
```

### Expected behavior

No client-side error

### To Reproduce

Use the `ErrorBoundary` component anywhere in the app. Only happens in production builds, webpack is used for both (no `--turbo` flag in dev).

Using Next.js 14.0.4 with standalone output and `outputFileTracingRoot` set (monorepo setup likely not required to reproduce).

### Possible Solution

Could be issues with treeshaking/minification, that's what 1.20.10 introduced. Downgrading to 1.20.9 resolves the issue.

manudeli:
I reproduced it in my next.js repo I found bug too. I made fixed new version in v1.23.2
but, Could you check whether v1.23.2 fix your issue please too?

manudeli:
If this version resolve your issue, close this issue please. Thanks for your help!

manudeli:
@minsoo-web @okinawaa Check it out please.

PR #584, opened 2024-01-10
fix(tsup): remove treeshaking, minification

manudeli:
fix: #583

# Overview

@relsunkaev

## PR Checklist


minsoo-web:
I wonder what the problem is that you deleted it!

minsoo-web:
I wonder what the problem is that you deleted it!

PR #615, opened 2024-01-17
fix(react): new internal api `SuspensiveError`

### LINKED ISSUE #598: [Feature]: assert method of useErrorBoundary, assert
### Package Scope

@suspensive/react

### Description

```tsx

errorBoundary.assert(Math.random() > 0.5, () => new CustomError('Math.random() should be greater than 0.5'))
```

### Possible Solution

_No response_

### etc.

_No response_

manudeli:
fix #598 

new internal api: `SuspensiveError` to assert each hooks position for ErrorBoundary, ErrorBoundaryGroup
SuspensiveError shouldn't be caught by ErrorBoundary. so I updated implementation of `<ErrorBoundary/>`

## PR Checklist


manudeli:
@minsoo-web @okinawaa Check this out please if can

minsoo-web (commented):
How about an interface like this?

```ts
ThrowError.only(someCondition, "error message")

ThrowError.if(someCondition, "error message")
```

Since the error is caused by a constraint, I thought I'd give it the nuance of a control statement!

manudeli:
> How about an interface like this?
> 
> ```ts
> ThrowError.only(someCondition, "error message")
> 
> ThrowError.if(someCondition, "error message")
> ```
> 
> Since the error is caused by a constraint, I thought I'd give it the nuance of a control statement!

Could you ThrowError's implementation more? if ThrowError extending Error, I think ThrowError is not proper Error name. because Error extending original JavaScript Error should be expressed more specific situation should be thrown. but name ThrowError is not specific. so describing more implementation for ThrowError is needed for me

minsoo-web:
>Could you share ThrowError's implementation more? if ThrowError extending Error, I think name ThrowError should be changed properly. because Error extending original JavaScript Error should be expressed more specific situation should be thrown. but name ThrowError is not specific. so describing more implementation for ThrowError is needed for me  
> We can imagine ErrorBoundary using shouldCatch prop with ThrowError

I agree with you, and the reason I left a comment (to change it to `ThrowError`) is that unlike Error in js, it only throws an error when **certain conditions are met**. 
To be precise, I didn't say that `AssersionError` should be changed to `ThrowError`, but rather that **it should be named assert to give it a little more nuance** that it only throws an error when **certain conditions** are met, like `if` or `only`.

I agree with keeping it as AssersionError.

```
// Before
AssersionError.assert(someCondition, "errorMessage")

// suggestion
AssersionError.if(someCondition, "errorMessage")
AssersionError.only(someCondition, "errorMessage")
```

`assert` is fine for now, but I'd appreciate it if you could take another look at naming things like `if` and `only`!

manudeli:
@minsoo-web I understood your intention, but if we throw error if condition. We can't assert something.
AssertionError should guarantee some condition.

Maybe you want to suggest below one. but assert is for asserting some condition POSITIVELY not negatively
```tsx
AssertionError.ifNot(condition, errorMessage)
if(!condition){ // This should use negative condition to assert condition
  throw new AssertionError.ifNot(errorMessage)
}

condition === true // true - condition will be always true TypeScriptly
```

```tsx
const Page = () => {
  const params = useParams<{adId: string}>()
  params.adId // string | undefined

  AssertionError.if(typeof params.adId !== 'string') // we CAN'T assert parameter adId as string POSITIVELY
  AssertionError.ifNot(typeof params.adId === 'string') // we CAN'T assert parameter adId as string POSITIVELY, ifNot have negative meaning
  const query = useSuspenseQuery({
    queryKey: ['ads', adId]
    queryFn: () => {...}
  })

  query.data // TData
}
```

So I thought we should do naming this method should use condition positively
then how about naming this method as `guarantee` as alternative?
```tsx
AssertionError.assert(condition, errorMessage) // we CAN use condition as POSITIVELY
AssertionError.guarantee(condition, errorMessage) // this is too.

condition === true // true - condition will be always true TypeScriptly
```

```tsx
const Page = () => {
  const params = useParams<{adId: string}>()
  params.adId // string | undefined

  AssertionError.assert(typeof params.adId === 'string') // we can assert parameter adId as string positively
  AssertionError.guarantee(typeof params.adId === 'string') // we can guarantee parameter adId as string positively
  const query = useSuspenseQuery({
    queryKey: ['ads', adId], // we don't need to use enabled and guarantee string params.adId TypeScriptly and Runtimely
    queryFn: () => {...}
  })

  query.data // TData
}
```

manudeli:
@minsoo-web When I first make this AssertionError, I wanted to exposure it as public api. but AssertionError is not too important feature to make releasing v2 be delayed. so first I'll make this feature as internal feature for asserting that hook is in children of @suspensive/react's useErrorBoundary, useErrorBoundaryFallbackProps, useErrorBoundaryGroup. I'll request review again

manudeli:
@minsoo-web there is no public api's change. Please review this Pull Request again

manudeli:
@minsoo-web Sorry to call you frequently,, but I made change of this feature. new internal api SuspensiveError can't be caught by ErrorBoundary of @suspensive/react. so We can correctly assert that useErrorBoundaryFallbackProps should be in errorBoundaryFallbackProps.fallback

minsoo-web (approved):
COOL!!! 
I don't think it needs to be external, and I've checked all the comments and things you said, let's go! 🚀🚀🚀

PR #656, opened 2024-01-26
fix(react): add getServerSnapshot of use-sync-external-store in for syncDevMode in SSR environment

### LINKED ISSUE #655: [Bug]: `SuspensiveProvider` breaks Next.js 14.1 dev mode.
### Package Scope

@suspensive/react

### Bug description

I am using Next.js 14.1 with the "pages router" and React 18, and I started having issues when I recently upgraded from `@suspensive/react` `1.20.8` to `1.25.2`

The production build of my app still worked; but in dev mode I could not get past the error "Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering."

The only solution I could find was to remove the `SuspensiveProvider` from my `_app.tsx` file.

image

### Expected behavior

With an earlier version (although I can't be sure which version precisely was installed, because I had it as `^1.20.8` in my `package.json`), I had no problems in dev or production mode.

### To Reproduce

_No response_

### Possible Solution

_No response_

### etc.

_No response_

manudeli:
fix #655 

# Overview

@lunelson Thanks Lu Nelson we could catch this bug quickly because of your reporting issue.
I added you as co-author for this Pull Request. Thanks

## PR Checklist

### DIFF
--- a/.changeset/happy-scissors-wait.md
+++ b/.changeset/happy-scissors-wait.md
@@ -1,0 +1,5 @@
+---
+"@suspensive/react": patch
+---
+
+fix(react): add getServerSnapshot in for syncDevMode in SSR environment
--- a/examples/next-streaming-react-query/package.json
+++ b/examples/next-streaming-react-query/package.json
@@ -10,7 +10,6 @@
   },
   "dependencies": {
     "@suspensive/react": "workspace:*",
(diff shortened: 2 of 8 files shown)


lunelson:
Thanks for the very fast response to this issue @manudeli !!

PR #709, opened 2024-02-07
fix(react): define default prop of each components(Delay, Suspense)

ssi02014:
# Overview

I was looking at the `Delay` component in `@suspensive/react` and it seems that if we utilize the generic type of `PropsWithChildren`, we can remove the unnecessary `OmitKeyOf` utility type.

```ts
type PropsWithChildren<P = unknown> = P & { children?: ReactNode | undefined };
```

Also, I added test code for the fallback of the Delay component. 

## PR Checklist

### DIFF
--- a/.changeset/swift-foxes-exist.md
+++ b/.changeset/swift-foxes-exist.md
@@ -2,4 +2,4 @@
 "@suspensive/react": patch
 ---
 
-test(react): add test code for Delay fallback
+fix(react): define default prop of each components
--- a/.changeset/swift-foxes-exist.md
+++ b/.changeset/swift-foxes-exist.md
@@ -1,0 +1,5 @@
+---
+"@suspensive/react": patch
+---
+
+test(react): add test code for Delay fallback
--- a/packages/react/src/Delay.spec.tsx
+++ b/packages/react/src/Delay.spec.tsx
@@ -1,15 +1,16 @@
 import { CustomError, TEXT } from '@suspensive/test-utils'
 import { render, screen, waitFor } from '@testing-library/react'
 import ms from 'ms'
-import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { Delay } from './Delay'
 import { Delay_ms_prop_should_be_greater_than_or_equal_to_0, SuspensiveError } from './models/SuspensiveError'
 
-beforeAll(() => {
+beforeEach(() => {
   vi.useFakeTimers({ shouldAdvanceTime: true })
 })
 
-afterAll(() => {
+afterEach(() => {
+  vi.runOnlyPendingTimers()
   vi.useRealTimers()
 })
 
@@ -35,12 +36,9 @@ describe('<Delay/>', () => {
         {TEXT}
       </Delay>
     )
-
-    vi.advanceTimersByTime(ms('0.5s'))
-
     expect(screen.queryByRole('paragraph')).toBeInTheDocument()
 
-    vi.advanceTimersByTime(ms('0.5s'))
+    vi.advanceTimersByTime(ms('1s'))
 
     await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument())
   })
--- a/packages/react/src/Delay.spec.tsx
+++ b/packages/react/src/Delay.spec.tsx
@@ -1,9 +1,17 @@
 import { CustomError, TEXT } from '@suspensive/test-utils'
 import { render, screen, waitFor } from '@testing-library/react'
 import ms from 'ms'
-import { describe, expect, it } from 'vitest'
+import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest'
 import { Delay } from './Delay'
 import { Delay_ms_prop_should_be_greater_than_or_equal_to_0, SuspensiveError } from './models/SuspensiveError'
+
+beforeAll(() => {
+  vi.useFakeTimers({ shouldAdvanceTime: true })
+})
+
+afterAll(() => {
+  vi.useRealTimers()
+})
 
 describe('<Delay/>', () => {
   it('should render the children after the delay', async () => {
@@ -18,7 +26,23 @@ describe('<Delay/>', () => {
   })
   it('should accept 0 for ms prop', async () => {
     render(<Delay ms={0}>{TEXT}</Delay>)
-    await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument(), { timeout: 1000 })
+
+    expect(screen.queryByText(TEXT)).toBeInTheDocument()
+  })
+  it('should render fallback content initially and then the actual text after the delay', async () => {
+    render(
+      <Delay ms={ms('1s')} fallback={<p role="paragraph">fallback</p>}>
+        {TEXT}
+      </Delay>
+    )
+
+    vi.advanceTimersByTime(ms('0.5s'))
+
+    expect(screen.queryByRole('paragraph')).toBeInTheDocument()
+
+    vi.advanceTimersByTime(ms('0.5s'))
+
+    await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument())
   })
   it('should throw SuspensiveError if negative number is passed as ms prop', () => {
     expect(() => render(<Delay ms={-1}>{TEXT}</Delay>)).toThrow(Delay_ms_prop_should_be_greater_than_or_equal_to_0)
--- a/packages/react/src/Delay.spec.tsx
+++ b/packages/react/src/Delay.spec.tsx
@@ -24,7 +24,7 @@ describe('<Delay/>', () => {
     render(<Delay>{TEXT}</Delay>)
     expect(screen.queryByText(TEXT)).toBeInTheDocument()
   })
-  it('should accept 0 for ms prop', async () => {
+  it('should accept 0 for ms prop', () => {
     render(<Delay ms={0}>{TEXT}</Delay>)
 
     expect(screen.queryByText(TEXT)).toBeInTheDocument()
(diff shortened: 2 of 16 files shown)


packages/react/src/contexts/DefaultOptionsContexts.ts
```diff
@@ -2,8 +2,9 @@ import { createContext } from 'react'
 import type { DelayProps, SuspenseProps } from '..'
 import type { OmitKeyOf } from '../utility-types'
 
-export const DelayDefaultPropsContext = createContext<OmitKeyOf<DelayProps, 'children'>>({
+export const DelayDefaultPropsContext = createContext<DelayProps>({
   ms: undefined,
+  fallback: undefined,
```
ssi02014, line 7:
```ts
  
const fallback = typeof props.fallback === 'undefined' ? defaultProps.fallback : props.fallback
```

Since we use `defaultProps.fallback` as above, what do you think about defining fallback for clarity?

`SuspenseDefaultPropsContext` already has a fallback defined. 

```tsx
export const SuspenseDefaultPropsContext = createContext<OmitKeyOf<SuspenseProps, 'children' | 'devMode'>>({
  fallback: undefined,
  clientOnly: undefined,
})
```

packages/react/src/Delay.spec.tsx
```diff
@@ -1,10 +1,18 @@
...
 import { Delay } from './Delay'
 import { Delay_ms_prop_should_be_greater_than_or_equal_to_0, SuspensiveError } from './models/SuspensiveError'
 
+beforeAll(() => {
+  vi.useFakeTimers({ shouldAdvanceTime: true })
+})
+
+afterAll(() => {
+  vi.useRealTimers()
+})
```
ssi02014, line 14: ### Related
https://github.com/testing-library/react-testing-library/issues/1198

packages/react/src/contexts/DefaultOptionsContexts.ts
```diff
@@ -2,8 +2,9 @@ import { createContext } from 'react'
 import type { DelayProps, SuspenseProps } from '..'
 import type { OmitKeyOf } from '../utility-types'
 
-export const DelayDefaultPropsContext = createContext<OmitKeyOf<DelayProps, 'children'>>({
+export const DelayDefaultPropsContext = createContext<DelayProps>({
```
manudeli, line 5: > In my opinion, if this is named for meaning props of Delay, because Delay have children prop, we should use PropsWithChildren in DelayProps internally
> 
> We can check ErrorBoundaryProps contain type PropsWithChildren internally
https://github.com/bvaughn/react-error-boundary/blob/master/src/types.ts#L18C33-L26

Because of this reason, we had to omit children of each components(Delay, Suspense) intentionally. because we didn't want library users to use default prop for children prop. children prop should be set by use case each time

packages/react/src/contexts/DefaultOptionsContexts.ts
manudeli: Cool!

packages/react/src/Delay.spec.tsx
```diff
@@ -16,9 +24,25 @@ describe('<Delay/>', () => {
...
+      </Delay>
+    )
+
+    vi.advanceTimersByTime(ms('0.5s'))
+
+    expect(screen.queryByRole('paragraph')).toBeInTheDocument()
+
+    vi.advanceTimersByTime(ms('0.5s'))
+
+    await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument())
```
manudeli, line 45: Cool~!

packages/react/src/Delay.tsx
```diff
@@ -3,15 +3,15 @@ import { DelayDefaultPropsContext } from './contexts'
...
+export interface DelayProps {
   ms?: number
   /**
    * @experimental This is experimental feature.
    */
   fallback?: ReactNode
 }
 
-export const Delay = (props: DelayProps) => {
+export const Delay = (props: PropsWithChildren<DelayProps>) => {
```
manudeli, line 14:
```suggestion
export const Delay = (props: DelayProps) => {
```

In my opinion, if this is named for meaning props of Delay, because Delay have children prop, we should use PropsWithChildren in DelayProps internally

We can check ErrorBoundaryProps contain type PropsWithChildren internally
https://github.com/bvaughn/react-error-boundary/blob/master/src/types.ts#L18C33-L26

packages/react/src/Delay.tsx
```diff
@@ -3,15 +3,15 @@ import { DelayDefaultPropsContext } from './contexts'
 import { useTimeout } from './hooks'
 import { Delay_ms_prop_should_be_greater_than_or_equal_to_0, SuspensiveError } from './models/SuspensiveError'
 
-export interface DelayProps extends PropsWithChildren {
+export interface DelayProps {
```
manudeli, line 6:
```suggestion
export interface DelayProps extends PropsWithChildren {
```

packages/react/src/contexts/DefaultOptionsContexts.ts
manudeli, line 5:
```suggestion
export const DelayDefaultPropsContext = createContext<OmitKeyOf<DelayProps, 'children'>>({
```

packages/react/src/wrap.tsx
```diff
@@ -130,7 +130,7 @@ export const withErrorBoundary = <TProps extends ComponentProps<ComponentType> =
  */
 export const withDelay = <TProps extends ComponentProps<ComponentType> = Record<string, never>>(
   component: ComponentType<TProps>,
-  delayProps: OmitKeyOf<DelayProps, 'children'> = {}
+  delayProps: DelayProps = {}
```
manudeli, line 133:
```suggestion
  delayProps: OmitKeyOf<DelayProps, 'children'> = {}
```

manudeli (changes requested):

.changeset/swift-foxes-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"@suspensive/react": patch
+---
+
+test(react): add test code for Delay fallback
```
manudeli, line 5:
```suggestion
fix(react): define default prop of each components
```

packages/react/src/contexts/DefaultOptionsContexts.ts
ssi02014: @manudeli Aha I understand the intent, thanks for the quick feedback  🙏

packages/react/src/Delay.spec.tsx
manudeli, line 45: How about strictly by timeout option of waitFor?

```suggestion
    await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument(), {
      timeout: ms('1s'),
    })
```

packages/react/src/Delay.spec.tsx
```diff
@@ -16,9 +24,25 @@ describe('<Delay/>', () => {
...
+      <Delay ms={ms('1s')} fallback={<p role="paragraph">fallback</p>}>
+        {TEXT}
+      </Delay>
+    )
+
+    vi.advanceTimersByTime(ms('0.5s'))
+
+    expect(screen.queryByRole('paragraph')).toBeInTheDocument()
+
+    vi.advanceTimersByTime(ms('0.5s'))
```
manudeli, line 43:
```suggestion
    expect(screen.queryByRole('paragraph')).toBeInTheDocument()
    vi.advanceTimersByTime(ms('1s'))
```

packages/react/src/Delay.spec.tsx
ssi02014: @manudeli 
I know that option doesn't make much sense if we use a `fake timer`, but do you think it would be better to add it?  

```tsx
  it('should render fallback content initially and then the actual text after the delay', async () => {
    render(/* ... */)
    // vi.advanceTimersByTime(ms('0.5s'))

    expect(screen.queryByRole('paragraph')).toBeInTheDocument()

    // vi.advanceTimersByTime(ms('0.5s'))

    await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument(), { timeout: 5000 })
  })
```
In the above case, the timeout of waitFor is valid. 

```tsx
  it('should render fallback content initially and then the actual text after the delay', async () => {
    render(/* ... */)
    // vi.advanceTimersByTime(ms('0.5s'))

    expect(screen.queryByRole('paragraph')).toBeInTheDocument()

    // vi.advanceTimersByTime(ms('0.5s'))

    await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument()) // 
  })
```

However, if we don't use fake time as shown above and don't specify the timeout option, the test will fail because, as you said, the default value of timeout is 1000ms.

```tsx
  it('should render fallback content initially and then the actual text after the delay', async () => {
    render(/* ... */)

    vi.advanceTimersByTime(ms('0.5s'))

    expect(screen.queryByRole('paragraph')).toBeInTheDocument()

    vi.advanceTimersByTime(ms('0.5s'))

    await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument())
  })
```
If we use a fake timer, we don't necessarily need the timeout option - the above test will work fine.

ssi02014: @manudeli Um.. I couldn't get an accurate test for timeout when I used `ms()`, so I entered it as a number.

ssi02014: I made the time value large for testing purposes. 🙏

manudeli (approved):
Thanks! In my opinion, using timeout option of waitFor is optional thing to approve this Pull Request. I think I need to study more about fake timer

packages/react/src/Delay.spec.tsx
ssi02014: Fix done! 

Additionally, I followed the guidelines in `testing-library`. The changes are as follows

```ts
beforeEach(() => {
  vi.useFakeTimers({ shouldAdvanceTime: true })
})

afterEach(() => {
  vi.runOnlyPendingTimers()
  vi.useRealTimers()
})
```

### Related
https://testing-library.com/docs/using-fake-timers/

ssi02014:
@manudeli Thanks! 👍

packages/react/src/Delay.spec.tsx
ssi02014: https://stackoverflow.com/questions/65970116/react-testing-library-waitfor-not-working-with-settimeout/65970771#65970771?newreg=dcb8efd321ab40cea4e0e0646d9b6934

PR #711, opened 2024-02-08
fix(react-query): useSuspenseQueries type error on select

Kanary159357:
# Overview

Fix https://github.com/suspensive/react/issues/710

## PR Checklist


packages/react-query/src/useSuspenseQueries.test-d.ts
```diff
@@ -3,9 +3,9 @@ import { expectTypeOf } from 'vitest'
 import { useSuspenseQueries } from '../dist'
 
 const queryKey = ['key'] as const
-const queryFn = () => 'response' as const
+const queryFn = () => ({ text: 'response' }) as const
 const boolean = Math.random() > 0.5
-const select = () => 'selected' as const
+const select = (arg: ReturnType<typeof queryFn>) => arg.text
```
manudeli, line 8:
```suggestion
const select = (data: Awaited<ReturnType<typeof queryFn>>) => data.text
```

packages/react-query/src/useSuspenseQueries.ts
```diff
@@ -31,11 +31,11 @@ interface ToInferWithEnabledOption<TQueryFnData, TQueryKey extends QueryKey, TEn
 }
 interface ToInferWithSelectOption<TQueryFnData, TData, TQueryKey extends QueryKey>
   extends ToInferOption<TQueryFnData, TQueryKey> {
-  select: (data: unknown) => TData
+  select: (data: any) => TData
```
manudeli, line 34: In my opinion, we should see useQueries implementation of TanStack React Query **v4**. because @suspensive/react-query target tanstack query v4. and they used any too. so we follow it.

https://github.com/TanStack/query/blob/v4/packages/react-query/src/useQueries.ts#L57

manudeli (approved):
We were missing this case(useSuspenseQueries with select) and I will create another PR to add test cases to guarantee that useSuspenseQuries with select is working. Thanks for your suggestion!

PR #731, opened 2024-02-14
feat(utility-types): add isNullable utility type

ssi02014:
# Overview

This is a minor suggestion! 🙏

How about managing types that have `null` as a union type with an `isNullable` utility type?

These types are meaningful and could be reused enough in the future. 

However, this task may be subject to personal preferences. **The opinion of the maintainer is important...!!💪**

In the case of `vitest`, it also includes `undefined` as an `isNullable` type, but we don't seem to need it.

```tsx
// vitest
type Nullable<T> = T | null | undefined;
```

## PR Checklist

### DIFF
--- a/packages/react/src/ErrorBoundary.tsx
+++ b/packages/react/src/ErrorBoundary.tsx
@@ -20,7 +20,7 @@ import {
   useErrorBoundaryFallbackProps_this_hook_should_be_called_in_ErrorBoundary_props_fallback,
   useErrorBoundary_this_hook_should_be_called_in_ErrorBoundary_props_children,
 } from './models/SuspensiveError'
-import type { ConstructorType, PropsWithDevMode } from './utility-types'
+import type { ConstructorType, Nullable, PropsWithDevMode } from './utility-types'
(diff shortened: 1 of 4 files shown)


packages/react/src/ErrorBoundary.tsx
```diff
@@ -187,7 +187,7 @@ export const ErrorBoundary = Object.assign(
   }
 )
 
-const ErrorBoundaryContext = createContext<({ reset: () => void } & ErrorBoundaryState) | null>(null)
+const ErrorBoundaryContext = createContext<Nullable<{ reset: () => void } & ErrorBoundaryState>>(null)
```
manudeli, line 190: In my opinion, this case is improvement than before. because we don't need to care parentheses to make type as nullable.

manudeli (approved):
Thanks 👍

minsoo-web:
Sorry for the late acknowledgement, your contribution is greatly appreciated! 
Thanks for the awesome PR!!!

ssi02014:
@manudeli @minsoo-web Thanks!! 🙏🎉

PR #840, opened 2024-04-08
docs(react-query): add more why we support this with @tanstack/react-query v4

### LINKED ISSUE #831: [Feature]: Merging interfaces of @tanstack/react-query and @suspensive/react into @suspensive/react-query
### Package Scope

@suspensive/react-query

### Description

Thank you for making great open source.
the number of apis provided by suspensive/react-query is increasing.

Tanstack Query is great, but it's unfortunate that there are compatibility issues.
I think suspense/react-query seem to aim to solve this problem. 

Couldn't it be provided to only use suspense/react-query?
I think it would be very convenient from the user's perspective.

### Possible Solution

_No response_

### etc.

_No response_

manudeli:
close #831 

@sonsurim @tooooo1 @sungh0lim @minsoo-web 

# Overview

This improve motivation and guide to use @suspensive/react-query with @tanstack/react-query v4, not v5

## PR Checklist


manudeli:
@minsoo-web merge first, improve gradually

PR #865, opened 2024-04-25
chore: add vitest globals tsconfig option

ssi02014:
# Overview

@manudeli  hello! 👋
It seems like it's been a while since I've created a pull request!

I was recently working on a migration from `jest` to `vitest` and came across the `globals` option.

vitest doesn't provide a global API by default, but for the `@suspensive/react` package, we explicitly enable the `globals` option.

So, by enabling the `globals` option, we don't need to import functions like `describe`, `it`, and `expect` in every file.

However, to apply it with typescript, you need to add the `vitest/globals` option to `tsconfig`.

We expect this will eliminate the need to import them per file and improve the developer experience !! 🤗

https://vitest.dev/config/#globals

What do you think of this work? I'd love to hear your thoughts!  🙏

## PR Checklist


manudeli (approved):
> What do you think of this work? I'd love to hear your thoughts! 🙏

Thanks! In my opinion, I guess this will make contributors can make test cases more easily

Issue #528, opened 2023-12-31
[Feature]: Determine what is `clientOnly prop of <Suspense/>`, `<ClientOnly/>`.


manudeli:
### Package Scope

@suspensive/react

### Possible Solution

1. Current implementation
2. useSyncExternalStore: https://twitter.com/TkDodo/status/1741068994981826947?t=XmG17etMUL2m0JFim03vqw&s=19
3. Global variable with initial state: https://baek2back.gitbook.io/fundamental/hooks/usehydrated

Collection50:
Hello, @manudeli 
I am very inspired by your activities.
`<ClientOnly/>` looks useful, but I don't think it's been worked on yet.
Can I be assigned this issue?

manudeli:
@Collection50 Thank you for your interest in Suspensive. If you submit it as a Pull Request, I think we can provide it as an experimental feature.

But, There are two reasons why we haven't implemented the ClientOnly component yet.

1. This is because the clear usage of the ClientOnly component has not yet been determined.

If you have a use case in mind, please suggest it.

2. The naming of ClientOnly may conflict with [clientOnly prop of Suspense](https://suspensive.org/docs/react/Suspense#avoid-server-side-rendering-clientonly). However, this may confuse library users because the internal implementations are different.

clientOnly in https://github.com/toss/suspensive/blob/main/packages/react/src/hooks/useIsClient.ts is handled by useIsomorphicLayoutEffect. I think it could be a different implementation than useSyncExternalStore in this issue and this could be the problem.

It would be nice to provide a unified implementation of the ClientOnly component and the clientOnly prop of Suspense in the same way.

manudeli:
@Collection50 Here is another implementation of ClientOnly: https://baek2back.gitbook.io/fundamental/hooks/usehydrated

1. Current implemantation: This way use component state to check if component self mounted.
2. Global Variable: This way use global variable as initial state of component to check if page already hydrated
3. useSyncExternalStore: This way use React 18 hook to check server or client

Collection50:
@manudeli 
I think [`useSyncExternalStore`](https://github.com/adobe/react-spectrum/blob/4e3af379e569faac3b374e0ed5a98b2a19cd92c3/packages/%40react-aria/ssr/src/SSRProvider.tsx#L190) is better way because it's `react` hook.

As you said, I was actually thinking about where to use `<ClientOnly/>`.
[...]

PR #964, opened 2024-06-24
fix(react): add `<ClientOnly/>` internally

Collection50:
related #528 

# Overview

Implemented the `ClientOnly` component.

## PR Checklist


packages/react/src/components/useIsClient.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+function getServerSnapshot() {
+  return true
+}
+
+// hook interface instead of useIsClient in hooks
+export function useIsClient() {
+  return useSyncExternalStore(() => noop, getSnapshot, getServerSnapshot)
+}
```
manudeli, line 15: This change is for next stage. Use [as-is useIsClient](https://github.com/toss/suspensive/blob/main/packages/react/src/hooks/useIsClient.ts) please first for current user.

packages/react/src/components/ClientOnly.tsx
```diff
@@ -0,0 +1,8 @@
+import { type SuspenseProps } from 'react'
+import { useIsClient } from './useIsClient'
+
+export const ClientOnly = ({ children, fallback }: SuspenseProps) => {
+  const isClient = useIsClient()
+
+  return isClient ? <>{children}</> : <>{fallback}</>
+}
```
manudeli, line 8: 1. Shortly if we can please
2. This is ClientOnly. so make type ClientOnlyProps please

```suggestion
export const ClientOnly = ({ children, fallback }: ClientOnlyProps) => <>{useIsClient() ? children : fallback}</>
```

packages/react/src/Suspense.tsx
```diff
@@ -1,10 +1,13 @@
...
-import { useIsClient } from './hooks'
 import type { PropsWithDevMode } from './utility-types'
 
-const SuspenseClientOnly = (props: ReactSuspenseProps) =>
-  useIsClient() ? <ReactSuspense {...props} /> : <>{props.fallback}</>
+const SuspenseClientOnly = (props: ReactSuspenseProps) => (
+  <ClientOnly fallback={props.fallback}>
+    <ReactSuspense {...props} />
+  </ClientOnly>
+)
```
manudeli, line 10: 👍

manudeli (changes requested):

packages/react/src/components/ClientOnly.tsx
Collection50: I think most of the time `<ClientOnly/>` be used as the HOC, so it need a `children`.
What do you think about using it as?

```ts
interface ClienOnlyProp {
  children: ReactNode:
  fallback?: ReactNode
}
```

manudeli: Cool! That's good for me

packages/react/src/components/useIsClient.ts
Collection50: How about this?

```ts
import React, { useState } from 'react'
import { noop } from '../utils'
import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect'

export const useIsClient = () => {
  const [isClient, setIsClient] = useState(false)

  useIsomorphicLayoutEffect(() => {
    setIsClient(true)
  }, [])

  if (typeof React['useSyncExternalStore'] === 'function') {
    return React['useSyncExternalStore'](
      () => noop,
      () => true,
      () => false
    )
  }

  return isClient
}
```

Collection50: This is a variant of the existing `useIsClient` and does not change the interface of the current `Suspensive` user; rather, the migration to the next stage is **much more acceptable**.

manudeli: > This is a variant of the existing useIsClient and does not change the interface of the current Suspensive user; rather, the migration to the next stage is much more acceptable.

I agree your opinion. but we should change our feature more gradually. I want to check this feature change as beta version. use original useIsClient please. In my opinion, This pull request should be just only for adding ClientOnly component. not improving Suspense clientOnly.

Collection50: Thank you for your opinion. I'll reflect it.

manudeli: Thanks for accepting my opinion. I also accept your opinion if we can too. but we should be more careful for this library users. 👍

Collection50: I'm just grateful to be able to contribute to a great library, with a developer I admire. If there's anything else I can contribute to, I'd love to.

packages/react/src/components/ClientOnly.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+    render(
+      <ClientOnly fallback={<div>Loading...</div>}>
+        <div>Client Content</div>
+      </ClientOnly>
+    )
+
+    expect(screen.getByText('Loading...')).toBeInTheDocument()
+    expect(screen.queryByText('Client Content')).not.toBeInTheDocument()
+  })
+})
```
manudeli, line 39: Cool test code!

manudeli (approved):
Thanks! @Collection50

manudeli:
@Collection50 resolve ci:type error please.

manudeli (approved):
@Collection50 Let's improve clientOnly gradually after this pull request merging

PR #966, opened 2024-06-24
fix(react): improved useTimeout and added usePreservedCallback

ssi02014:
# Overview

@manudeli Hello! 👋

Currently, `useTimeout` hook is written to call the changed function when the callback function changes.
However, in practice, we're not actually calling the changed function, even if the function changes in the middle.

This is because the only value we currently depend on for `useEffect` is `ms`. 
Adding `ref` to the dependency array doesn't solve this problem because it's not tracked.

These problems can be solved by using hooks like `usePreservedCallback` in `toss/slash`, which can preserve the reference of the callback function, while at the same time ensuring the function is up-to-date. 

- [usePreservedCallback](https://github.com/toss/slash/blob/main/packages/react/react/src/hooks/usePreservedCallback.ts)

Since the `ref` keeps updating during its lifecycle, when calling the function returned by `useCallback`, the latest function will be called by the `closure`.

## PR Checklist

### DIFF
--- a/packages/react/src/hooks/usePreservedCallback.ts
+++ b/packages/react/src/hooks/usePreservedCallback.ts
@@ -1,0 +1,14 @@
+import { useCallback, useRef } from 'react'
+import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect'
+
+export const usePreservedCallback = <T extends (...args: unknown[]) => unknown>(callback: T) => {
+  const callbackRef = useRef<T>(callback)
+
+  useIsomorphicLayoutEffect(() => {
+    callbackRef.current = callback
+  }, [callback])
+
+  return useCallback((...args: unknown[]) => {
+    return callbackRef.current(...args)
+  }, []) as T
+}
--- a/packages/react/src/hooks/useTimeout.spec.tsx
+++ b/packages/react/src/hooks/useTimeout.spec.tsx
@@ -16,6 +16,25 @@ describe('useTimeout', () => {
     expect(spy).toHaveBeenCalledTimes(1)
   })
 
+  it('should call the modified function when the given function is changed', async () => {
+    const timeoutMockFn1 = vi.fn()
+    const timeoutMockFn2 = vi.fn()
+
+    const result = renderHook(({ fn }) => useTimeout(fn, ms('0.1s')), {
+      initialProps: { fn: timeoutMockFn1 },
+    })
+
+    result.rerender({ fn: timeoutMockFn2 })
+
+    expect(timeoutMockFn1).toHaveBeenCalledTimes(0)
+    expect(timeoutMockFn2).toHaveBeenCalledTimes(0)
+
+    await sleep(ms('0.1s'))
+
+    expect(timeoutMockFn1).toHaveBeenCalledTimes(0)
+    expect(timeoutMockFn2).toHaveBeenCalledTimes(1)
+  })
+
   it('should not re-call callback received as argument even if component using this hook is rerendered', async () => {
     const TestComponent = () => {
       const [number, setNumber] = useState(0)
--- a/packages/react/src/hooks/useTimeout.spec.tsx
+++ b/packages/react/src/hooks/useTimeout.spec.tsx
@@ -20,11 +20,11 @@ describe('useTimeout', () => {
     const timeoutMockFn1 = vi.fn()
     const timeoutMockFn2 = vi.fn()
 
-    const result = renderHook(({ fn }) => useTimeout(fn, ms('0.1s')), {
+    const { rerender } = renderHook(({ fn }) => useTimeout(fn, ms('0.1s')), {
       initialProps: { fn: timeoutMockFn1 },
     })
 
-    result.rerender({ fn: timeoutMockFn2 })
+    rerender({ fn: timeoutMockFn2 })
 
     expect(timeoutMockFn1).toHaveBeenCalledTimes(0)
     expect(timeoutMockFn2).toHaveBeenCalledTimes(0)
--- a/packages/react/src/hooks/useTimeout.ts
+++ b/packages/react/src/hooks/useTimeout.ts
@@ -1,15 +1,11 @@
-import { useEffect, useRef } from 'react'
-import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect'
+import { useEffect } from 'react'
+import { usePreservedCallback } from './usePreservedCallback'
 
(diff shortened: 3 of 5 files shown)


packages/react/src/hooks/useTimeout.spec.tsx
```diff
@@ -16,6 +16,25 @@ describe('useTimeout', () => {
...
+    result.rerender({ fn: timeoutMockFn2 })
+
+    expect(timeoutMockFn1).toHaveBeenCalledTimes(0)
+    expect(timeoutMockFn2).toHaveBeenCalledTimes(0)
+
+    await sleep(ms('0.1s'))
+
+    expect(timeoutMockFn1).toHaveBeenCalledTimes(0)
+    expect(timeoutMockFn2).toHaveBeenCalledTimes(1)
+  })
```
ssi02014, line 36: If it's existing code, it won't call the changed function as intended, and the test will fail. 😭

packages/react/src/hooks/usePreservedCallback.ts
```diff
@@ -0,0 +1,14 @@
...
+  const callbackRef = useRef<T>(callback)
+
+  useIsomorphicLayoutEffect(() => {
+    callbackRef.current = callback
+  }, [callback])
+
+  return useCallback((...args: unknown[]) => {
+    return callbackRef.current(...args)
+  }, []) as T
+}
```
ssi02014, line 14: This is slightly different from `toss/slash's usePreservedCallback`. 

I used `useIsomorphicLayoutEffect` instead of `useEffect`, and changed the type to `unknown` instead of `any`.

packages/react/src/hooks/useTimeout.spec.tsx
```diff
@@ -16,6 +16,25 @@ describe('useTimeout', () => {
...
+    rerender({ fn: timeoutMockFn2 })
+
+    expect(timeoutMockFn1).toHaveBeenCalledTimes(0)
+    expect(timeoutMockFn2).toHaveBeenCalledTimes(0)
+
+    await sleep(ms('0.1s'))
+
+    expect(timeoutMockFn1).toHaveBeenCalledTimes(0)
+    expect(timeoutMockFn2).toHaveBeenCalledTimes(1)
+  })
```
ssi02014, line 36: If it's existing code, it won't call the changed function as intended, and the test will fail. 😭

packages/react/src/hooks/useTimeout.ts
```diff
@@ -1,15 +1,11 @@
...
+import { useEffect } from 'react'
+import { usePreservedCallback } from './usePreservedCallback'
 
 export const useTimeout = (fn: () => void, ms: number) => {
-  const fnRef = useRef(fn)
-
-  useIsomorphicLayoutEffect(() => {
-    fnRef.current = fn
-  }, [fn])
+  const preservedCallback = usePreservedCallback(fn)
```
manudeli, line 5: I think Components([Delay](https://github.com/toss/suspensive/blob/main/packages/react/src/Delay.tsx#L21)) using useTimeout should use useCallback too.

manudeli (changes requested):
1. @suspensive/test-utils should be changed like this implementation.
2. To make how to work be same between AS-IS and TO-BE, we should add useCallback in all components using useTimeout. There is no significant difference in components that use useTimeout, but it can be expected that performance will be slightly worse. So I'm confuse that this is really improvement. I need more reasons to merge this change

ssi02014:
@manudeli 

> To make how to work be same between AS-IS and TO-BE, we should add useCallback in all components using useTimeout. There is no significant difference in components that use useTimeout, but it can be expected that performance will be slightly worse. So I'm confuse that this is really improvement. I need more reasons to merge this change

I did this because what was already written wasn't working as intended 🤔
If you don't want to guarantee up-to-date functions, the code below should be removed because it's pointless.

```diff
import { useEffect, useRef } from 'react'
import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect'

export const useTimeout = (fn: () => void, ms: number) => {
-  const fnRef = useRef(fn)

-  useIsomorphicLayoutEffect(() => {
-    fnRef.current = fn
-  }, [fn])

  useEffect(() => {
-    const id = setTimeout(fnRef.current, ms)
+    const id = setTimeout(fn, ms)
    return () => clearTimeout(id)
  }, [ms])
}
```

packages/react/src/hooks/useTimeout.ts
manudeli: @ssi02014 I absolutely accept your opinion https://github.com/toss/suspensive/pull/966#issuecomment-2185931597 but anyway

I think Components([Delay](https://github.com/toss/suspensive/blob/main/packages/react/src/Delay.tsx#L21)) using useTimeout should use useCallback too. doesn't it right?

### TO-BE
```diff
import { type PropsWithChildren, type ReactNode, useContext, useState } from 'react'
import { DelayDefaultPropsContext } from './contexts'
import { useTimeout } from './hooks'
import { Message_Delay_ms_prop_should_be_greater_than_or_equal_to_0, SuspensiveError } from './models/SuspensiveError'

export interface DelayProps extends PropsWithChildren {
  ms?: number
  fallback?: ReactNode
}

export const Delay = (props: DelayProps) => {
  if (process.env.NODE_ENV === 'development') {
    if (typeof props.ms === 'number') {
      SuspensiveError.assert(props.ms >= 0, Message_Delay_ms_prop_should_be_greater_than_or_equal_to_0)
    }
  }
  const defaultProps = useContext(DelayDefaultPropsContext)
  const ms = props.ms ?? defaultProps.ms ?? 0

  const [isDelaying, setIsDelaying] = useState(ms > 0)
-  useTimeout(() => setIsDelaying(false), ms)
+  useTimeout(useCallback(() => setIsDelaying(false), []), ms)

  const fallback = typeof props.fallback === 'undefined' ? defaultProps.fallback : props.fallback
  return <>{isDelaying ? fallback : props.children}</>
}
if (process.env.NODE_ENV === 'development') {
  Delay.displayName = 'Delay'
}
```

ssi02014: @manudeli Oh no, `useTimeout` works the same way.
No need to change it! 👍

manudeli: Ah! I understood what you mean! Thanks

manudeli (approved):
Much better!

ssi02014:
@manudeli I recently took a deep dive into the `usePreservedCallback` hook. 
As a result of this reflection, I'm happy to contribute to a great library once again.

PR #978, opened 2024-06-26
feat(configs): add options for scripts

gwansikk:
# Overview

- Since the current project uses `configs/tsup`, I have added the configuration for scripts.
- Currently, this script is only used in the react-query package, but there is potential to use this config in other packages in the future.

## PR Checklist

### DIFF
--- a/configs/tsup/src/index.ts
+++ b/configs/tsup/src/index.ts
@@ -15,7 +15,7 @@ export const scriptOptions: Options = {
   clean: true,
   format: ['cjs'],
   target: ['chrome51', 'firefox53', 'edge18', 'safari11', 'ios11', 'opera38', 'es6', 'node14'],
-  entry: ['src/scripts/**/*.{ts,tsx}', '!**/*.{spec,test,test-d,bench}.*'],
+  entry: ['src/scripts/*.{ts,tsx}', '!**/*.{spec,test,test-d,bench}.*'],
   outDir: 'dist/scripts',
   sourcemap: true,
 }
--- a/configs/tsup/src/index.ts
+++ b/configs/tsup/src/index.ts
@@ -10,3 +10,12 @@ export const options: Options = {
   sourcemap: true,
   dts: true,
 }
+
+export const scriptOptions: Options = {
+  clean: true,
+  format: ['cjs'],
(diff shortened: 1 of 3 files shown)


configs/tsup/src/index.ts
```diff
@@ -10,3 +10,12 @@ export const options: Options = {
   sourcemap: true,
   dts: true,
 }
+
+export const scriptOptions: Options = {
+  clean: true,
+  format: ['cjs'],
+  target: ['chrome51', 'firefox53', 'edge18', 'safari11', 'ios11', 'opera38', 'es6', 'node14'],
+  entry: ['src/scripts/**/*.{ts,tsx}', '!**/*.{spec,test,test-d,bench}.*'],
```
manudeli, line 18: I think we should remove this `'src/scripts/**/*.{ts,tsx}'`. check out please!

```suggestion
  entry: ['src/scripts/*.{ts,tsx}', '!**/*.{spec,test,test-d,bench}.*'],
```

manudeli (changes requested):

configs/tsup/src/index.ts
gwansikk: > I think we should remove this `'src/scripts/**/*.{ts,tsx}'`. check this out please!

That's a good point! There's no need to build the parent directory (scripts/utils) that is only used by scripts and not by the user.

manudeli (approved):
Clean architecture! 🚀

manudeli:
@gwansikk Check this change

AS-IS: 33.9kB, 45 files https://www.npmjs.com/package/@suspensive/react-query/v/2.2.3
image

TO-BE: 23.4kB, 34 files https://www.npmjs.com/package/@suspensive/react-query/v/2.2.4
image

gwansikk:
> @gwansikk Check this change
> 
> AS-IS: 33.9kB, 45 files https://www.npmjs.com/package/@suspensive/react-query/v/2.2.3 image
> 
> TO-BE: 23.4kB, 34 files https://www.npmjs.com/package/@suspensive/react-query/v/2.2.4 image

That's right. Due to `entry: ['src/scripts/*.{ts,tsx}']` and `dts: false` in scriptOptions, the build files and `~.d.ts`, `~.d.cts` from utils have been removed, reducing the number of files.

- The code related to scripts is not intended for use by users or TypeScript, so `type declarations` have been disabled.

Issue #986, opened 2024-06-27
[Feature]: test code for cli


manudeli:
### Package Scope

@suspensive/react-query

### Description

- [ ] cli.spec.ts
- [ ] postinstall.spec.ts
- [x] commands.spec.ts
- [x] copy.spec.ts
- [ ] loadModule.spec.ts
- [x] package.spec.ts
- [x] switchVersion.spec.ts

PR #1008, opened 2024-06-29
test(react-query): add tests for commands, switchVersion, copy, package

gwansikk:
# Overview

add tests for cli

- commands
- copy
- package
- switchVersion

image

## PR Checklist


gwansikk:
@manudeli Do you have any good ideas for testing file reading, copying, and overwriting as in `copy.ts`?
I think using a library like [mock-fs](https://github.com/tschaub/mock-fs) along with mock functions would be a good approach.

manudeli:
> @manudeli Do you have any good ideas for testing file reading, copying, and overwriting as in `copy.ts`? I think using a library like [mock-fs](https://github.com/tschaub/mock-fs) along with mock functions would be a good approach.

Let's try it as devDeps at now!

manudeli:
@gwansikk one file by one file could be good way to do this! because this could make conflict if main branch updated.

gwansikk:
> @gwansikk one file by one file could be good way to do this! because making this test coverage at once could make conflict if main branch updated frequently.

I agree, I'll track the remaining work in #986. I think I will finish everything by today!

| [Feature]: test code for cli #986  |
| --- |
|image|

packages/react-query/src/scripts/cli.test.ts
gwansikk, line 1: Currently, we run the built files to check if they work correctly. As a result, the coverage for `cli.ts` is not collected and remains at 0%. We need to write test codes for `cli.ts` using mocking, but it is complex and challenging (we need a way to validate it without running the CLI with Node).

For now, I have temporarily added test codes to ensure a stable CLI deployment. I believe this will provide some preemptive protection for the CLI.

I spent some time trying various methods to test `cli.ts` without executing it with Node through mocking, but I couldn't find a good solution. If you know of any effective methods, please let me know!

manudeli (commented):
Could we use *.spec.ts file than *.test.ts? Do you have reason to choose *.test.ts?

packages/react-query/src/scripts/utils/copy.ts
```diff
@@ -6,7 +6,7 @@ const dir = resolve(__dirname, '..')
 export function copy(version: number) {
   const files = fs.readdirSync(dir)
 
-  if (!dir.includes('dist') || files.length === 0) {
+  if (files.length === 0) {
```
manudeli, line 9: If this change should be, we should add changeset file too.

gwansikk:
> Could we use *.spec.ts file than *.test.ts? Do you have reason to choose *.test.ts?

Oh, I'm sorry, my mistake..!

manudeli:
@gwansikk Could you add this changeset in this https://github.com/toss/suspensive/pull/1008#issuecomment-2198238225
because of https://github.com/toss/suspensive/pull/1008#discussion_r1661008535

gwansikk:
@manudeli There have been some changes..!

- I conducted the tests using vitest mocking instead of mock-fs.
- Testing if the CLI in `cli.ts` works well by executing it with Node was challenging. Therefore, I created `commands.ts` to separately handle the command actions for testing, and changed `cli.spec.ts` to test if it works correctly in an actual built environment. check this: [refactor(react-query): add commands file for cli actions](https://github.com/toss/suspensive/pull/1008/commits/8c986c6f34e5139c66d9e001e4e1e520226844b8)
- I wrote test codes for each utility using appropriate mocking: `commands.spec.ts`, `copy.spec.ts`, `package.spec.ts`, `switchVersion.spec.ts`.

manudeli:
> * I conducted the tests using vitest mocking instead of mock-fs.

Much better...! 

> * Testing if the CLI in `cli.ts` works well by executing it with Node was challenging. Therefore, I created `commands.ts` to separately handle the command actions for testing, and changed `cli.spec.ts` to test if it works correctly in an actual built environment. check this: [refactor(react-query): add commands file for cli actions](https://github.com/toss/suspensive/pull/1008/commits/8c986c6f34e5139c66d9e001e4e1e520226844b8)
> * I wrote test codes for each utility using appropriate mocking: `commands.spec.ts`, `copy.spec.ts`, `package.spec.ts`, `switchVersion.spec.ts`.
> 
> As a result
> 
> image

Coool! This will make stability to our library 💉

packages/react-query/src/scripts/cli.ts
```diff
@@ -23,54 +16,17 @@ program
...
-    if (suspensiveReactQueryVersion === tanStackReactQueryMajorVersion) {
-      console.log(`\nThe versions are compatible.`)
-    } else {
-      console.warn(
-        '\nThe version of @suspensive/react-query is not compatible with the current version of @tanstack/react-query.',
-        `\nPlease run 'npx suspensive-react-query switch ${suspensiveReactQueryVersion === '5' ? '4' : '5'}' to switch to the compatible version.`
-      )
-    }
-  })
+  .action(() => statusAction())
```
manudeli, line 19: Cool! testable code

.changeset/clever-steaks-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@suspensive/react-query": patch
+---
+
+test(react-query): add tests for commands, switchVersion, copy, package
```
manudeli, line 5: Cool, your first changeset! 🚀

manudeli (approved):
🚀🚀🚀🚀🚀 Let's go to stable cli 🚀🚀🚀🚀🚀

Issue #1000, opened 2024-06-28
[Feature]: More description for target version


manudeli:
### Package Scope

@suspensive/react-query

### Bug description

<img width="362" alt="image" src="https://github.com/toss/suspensive/assets/61593290/ccc758d2-b35a-4231-9366-107db583e319">

issued by @juno7803

manudeli:
@juno7803 Could you add what you expect?

juno7803:
I believe that the cli for @suspensive/react-query status clearly supports different versions of react-query for users. However, despite its clear functionality, I find the guidance provided by the CLI to be unclear. If it's alright with you, could I contribute by submitting a Pull Request? 😁

gwansikk:
> I believe that the cli for @suspensive/react-query status clearly supports different versions of react-query for users.
> 
> However, despite its clear functionality, I find the guidance provided by the CLI to be unclear. If it's alright with you, could I contribute by submitting a Pull Request? 😁

@juno7803  I agree, especially the `(5)` in `[@suspensive/react-query] v2.3.0 (5)` is not clear. It is difficult for users to understand what it means. Please revise it to provide sufficient guidance from the user's perspective.

manudeli:
## Suggestion
I made example what I expected.

### Suspensive React Query with TanStack Query v4

```text
Suspensive React Query status:
@suspensive/react-query@2.3.0 export @suspensive/react-query-4's interfaces
  - useSuspenseQuery
  - useSuspenseQueries
  - useSuspenseInfiniteQuery
  - queryOptions
  - SuspenseQuery
  - SuspenseQueries
  - SuspenseInfiniteQuery
  - QueryErrorBoundary
@tanstack/react-query@4.36.2
  - all interfaces
```


### Suspensive React Query with TanStack Query v5
```text
Suspensive React Query status:
@suspensive/react-query@2.3.0 export @suspensive/react-query-5's interfaces
  - SuspenseQuery
  - SuspenseQueries
  - SuspenseInfiniteQuery
  - QueryErrorBoundary
@tanstack/react-query@5.32.0
  - useSuspenseQuery
  - useSuspenseQueries
  - useSuspenseInfiniteQuery
  - queryOptions
  - all interfaces
```

gwansikk:
@juno7803 @manudeli 

https://github.com/toss/suspensive/issues/1000#issuecomment-2196488181

[...]

PR #1024, opened 2024-07-01
feat(react-query): more descriptions for cli status

gwansikk:
# Overview

I've added you as a co-author! @juno7803 
Thank you for your valuable feedback! Thanks to you, the explanation has become even better.

I have made the API display static for now. In the next version, I will make it dynamic to list and display the APIs. https://github.com/toss/suspensive/issues/1000#issuecomment-2197635287

| version 4 | version 5|
| --- | --- |
|image|image|

## PR Checklist

### DIFF
--- a/packages/react-query/src/scripts/utils/commands.ts
+++ b/packages/react-query/src/scripts/utils/commands.ts
@@ -3,6 +3,19 @@ import path from 'path'
 import { getPackageJson, getTanStackReactQueryPackageJson } from './package'
 import { switchVersion } from './switchVersion'
 
+export const version5APIs = ['SuspenseQuery', 'SuspenseQueries', 'SuspenseInfiniteQuery', 'QueryErrorBoundary']
+
+export const version4APIs = [
+  'useSuspenseQuery',
+  'useSuspenseQueries',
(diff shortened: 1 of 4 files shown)


packages/react-query/src/scripts/utils/commands.ts
```diff
@@ -3,6 +3,19 @@ import path from 'path'
 import { getPackageJson, getTanStackReactQueryPackageJson } from './package'
 import { switchVersion } from './switchVersion'
 
+export const version5APIs = ['SuspenseQuery', 'SuspenseQueries', 'SuspenseInfiniteQuery', 'QueryErrorBoundary']
```
gwansikk, line 6: I have temporarily declared the API list. In the next version, I will parse the TypeScript to list and display the APIs!

manudeli:
@gwansikk changeset please

packages/react-query/src/scripts/utils/commands.ts
manudeli: cool 👍 Fast resolve issue

manudeli (approved):
Much better 🚀🚀🚀🚀🚀

Issue #1051, opened 2024-07-08
[Bug]: GitHub fetching don't work. Fetch succeeds only intermittently.


manudeli:
### Bug description

https://suspensive.org/docs/contributors

<img width="1719" alt="image" src="https://github.com/toss/suspensive/assets/61593290/957f77fa-aaef-48ad-85f2-1148b0a4d052">

@saul-atomrigs Could I assign you this issue?

saul-atomrigs:
Sure, please! 

I suspect it might be due to github API rate limit..
I'll be able to have a closer look after company work hours, but please go ahead with the assignment! 👀

manudeli:
> Sure, please!
> 
> I suspect it might be due to github API rate limit.. I'll be able to have a closer look after company work hours, but please go ahead with the assignment! 👀

Thanks! I'll wait you for resolving this issue

manudeli:
@saul-atomrigs I close this issue by https://github.com/toss/suspensive/pull/1056

If you other opinion, open this issue again please

PR #1054, opened 2024-07-08
fix(suspensive.org): add defense code for github contributors fetching

saul-atomrigs:
# Overview

Issue: https://github.com/toss/suspensive/issues/1051

I was able to reproduce the bug only by the time this issue was posted. Afterwards, I tried clearing the cache, browsing in secret mode, and even in another macbook (which never visited the page before) and in several browsers (chrome, safari, opera, edge),  I couldn't reproduce the bug again... for now I added some defense code to where the error log pointed:

Captura de pantalla 2024-07-08 a las 8 02 06 p  m

I'll keep an eye on this issue. 

## PR Checklist

- [✅] I did below actions if need

### DIFF
--- a/docs/suspensive.org/src/lib/hooks/useFetchContributors.ts
+++ b/docs/suspensive.org/src/lib/hooks/useFetchContributors.ts
@@ -41,21 +41,36 @@ const useFetchContributors = () => {
           throw new Error('Failed to fetch contributors')
         }
         const data: GithubRepoContributor[] = (await response.json()) as GithubRepoContributor[]
-        if (Array.isArray(data) && data.length) {
-          const filteredContributors = data.filter((contributor) => {
-            const login = contributor.author.login
-            return !['github-actions[bot]', 'dependabot[bot]', 'renovate[bot]'].includes(login)
-          })
-          setResult({
-            data: filteredContributors,
-            error: undefined,
-            isError: false,
-            isLoading: false,
-            isSuccess: true,
-          })
+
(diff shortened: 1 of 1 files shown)


docs/suspensive.org/src/lib/hooks/useFetchContributors.ts
```diff
@@ -41,17 +41,19 @@ const useFetchContributors = () => {
...
+            return !['github-actions[bot]', 'dependabot[bot]', 'renovate[bot]'].includes(login)
+          })
+          setResult({
+            data: filteredContributors,
+            error: undefined,
+            isError: false,
+            isLoading: false,
+            isSuccess: true,
+          })
+        }
```
manudeli, line 56: Even if the data is not an array or the length is 0, the fetch success status must be expressed as state. Please modify isSuccess to be true so that the correct server status state can be displayed even if the length of the data is 0 or there is no data when the fetch is successful.

manudeli (changes requested):

docs/suspensive.org/src/lib/hooks/useFetchContributors.ts
saul-atomrigs: thank you, I've just pushed the change
It will keep `isSuccess` true for all 200 OK, even though the response data is empty or non-array
plus, I refactored the code for better exception messages

manudeli: Thanks!

Issue #433, opened 2023-12-05
[Feature]: awaitOptions


manudeli:
### Package Scope

@suspensive/react-await

### Description

awaitOptions

SEOKKAMONI:
@manudeli Hi! I'm interested in that issue, can you assign it to me?

manudeli:
Sure, I assigned you now

PR #1059, opened 2024-07-09
feat(react-await): add `awaitOptions`

SEOKKAMONI:
# Overview
resolved https://github.com/toss/suspensive/issues/433

- awaitOptions (experimental)
- add tests for awaitOptions, useAwait

## PR Checklist


docs/suspensive.org/src/pages/docs/react-await/awaitOptions.en.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+# awaitOptions
+
+<Callout type='experimental'>
+
+`<Await/>` is experimental feature, this interfaces could be changed
+
+</Callout>
+
+`awaitOptions` combines a key and a function to help manage the options in the current `<Await />` component.
```
SEOKKAMONI, line 11: If the interface changes in the future, we will also change the document. But I thought it was right to specify the <Await /> component to deliver explicitly as of now

packages/react-await/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export { Await, useAwait, awaitClient } from './Await'
+export { awaitOptions } from './awaitOptions'
 
-export type { AwaitProps } from './Await'
+export type { AwaitProps, AwaitOptions } from './Await'
```
SEOKKAMONI, line 4: With the development of 'awaitOptions', I thought that the AwaitOptions type should also be given access, so I allowed it as follows.

packages/react-await/src/awaitOpetions.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+    render(
+      <Suspense fallback={FALLBACK}>
+        <Await options={options}>{({ data }) => <>{data}</>}</Await>
+      </Suspense>
+    )
+
+    expect(await screen.findByText(TEXT)).toBeInTheDocument()
+  })
+
+  it('should ve used with useAwait', async () => {
```
SEOKKAMONI, line 22:
```suggestion
  it('should be used with useAwait', async () => {
```

packages/react-await/src/Await.tsx
```diff
@@ -4,7 +4,7 @@ import { hashKey } from './utils'
 
 export type Key = Tuple
 
-type AwaitOptions<TData, TKey extends Key> = {
+export type AwaitOptions<TData, TKey extends Key> = {
```
manudeli, line 7:
```suggestion
/**
 * @experimental This is experimental feature.
 */
export type AwaitOptions<TData, TKey extends Key> = {
```

manudeli (approved):
🚀🚀🚀

PR #1081, opened 2024-07-13
test(react): improve `ErrorBoundary` test code

jgjgill:
# Overview

hello. I've been looking at the `ErrorBoundary` test code and trying to improve it.

I did four things.

- Removed the comment because there is a test that resets to `resetKeys.length`.
- I thought it would be easier to understand the code if `shouldCatch` had all three cases, so I added `false`.
- `shouldCatch` related tests only have tests that fail (Not `Child`, Only `Parent`). I added tests that consider them all.
- For `user.click`, I added `await` by returning `Promise`.

Thanks for the great library. 🙇‍♂️

## PR Checklist


manudeli (approved):
Thanks 👍

packages/react/src/ErrorBoundary.spec.tsx
```diff
@@ -188,74 +186,112 @@ describe('<ErrorBoundary/>', () => {
...
+      parentCalledTimes: 1,
+      shouldCatch: CustomError,
+      errorText: `${ERROR_MESSAGE} of Parent`,
+    },
+    {
+      childCalledTimes: 1,
+      parentCalledTimes: 0,
+      shouldCatch: Error,
+      errorText: `${ERROR_MESSAGE} of Child`,
+    },
```
manudeli, line 201: Cool Case!!

packages/react/src/ErrorBoundary.spec.tsx
```diff
@@ -188,74 +186,112 @@ describe('<ErrorBoundary/>', () => {
...
+      parentCalledTimes: 1,
+      shouldCatch: (error: Error) => error instanceof CustomError,
+      errorText: `${ERROR_MESSAGE} of Parent`,
+    },
+    {
+      childCalledTimes: 1,
+      parentCalledTimes: 0,
+      shouldCatch: (error: Error) => error instanceof Error,
+      errorText: `${ERROR_MESSAGE} of Child`,
+    },
```
manudeli, line 240: Cool Case!!

Issue #1025, opened 2024-07-01
I have temporarily declared the API list. In the next version, I will parse the TypeScript to list and display the APIs!


manudeli:
I have temporarily declared the API list. In the next version, I will parse the TypeScript to list and display the APIs!

_Originally posted by @gwansikk in https://github.com/toss/suspensive/pull/1024#discussion_r1661294759_

manudeli:
@juno7803 This issue will be resolved by @gwansikk

manudeli:
@gwansikk Could you do this first please

gwansikk:
> @gwansikk Could you do this first please

I missed that. I'll get on it right away!

PR #1113, opened 2024-07-19
refactor(react-query): parse api list for status table from build files

gwansikk:
# Overview

- #1025 

## PR Checklist

### DIFF
--- a/packages/react-query/src/scripts/utils/loadModule.ts
+++ b/packages/react-query/src/scripts/utils/loadModule.ts
@@ -1,7 +1,7 @@
 export function loadModule<T>(name: string): T | undefined {
   try {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-    return require(name)
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    return require(name) as T
   } catch (e) {
     return undefined
   }
--- a/packages/react-query/src/scripts/utils/package.ts
+++ b/packages/react-query/src/scripts/utils/package.ts
@@ -1,3 +1,5 @@
+import fs from 'fs'
+import path from 'path'
 import { exit } from 'process'
 import { loadModule } from './loadModule'
 
@@ -48,4 +50,42 @@ export function getSuspensiveReactQueryPackageJson(targetVersion: string) {
   }
 
   return packageJson
+}
+
+export function getIndexFileContent() {
+  const basePath = path.resolve(__dirname, '../../dist').replace(/src/, '')
+
+  return fs.readFileSync(path.join(basePath, 'index.js'), 'utf-8') || ''
+}
+
+export function getTargetSuspensiveReactQueryVersion(): string {
+  const indexFileContent = getIndexFileContent()
+
+  return (RegExp(/@suspensive\/react-query-(\d+)/).exec(indexFileContent) || [])[1] || ''
+}
+
+export function getTargetSuspensiveReactQueryAPIs() {
+  const indexFileContent = getIndexFileContent()
+
+  const modules = [...indexFileContent.matchAll(/export \* from ['"](.+?)['"]/g)]
+  const results: string[] = []
+
+  for (const module of modules) {
+    const moduleName = module[1]
+    const moduleExports = loadModule<Record<string, unknown>>(moduleName)
+
+    if (moduleExports === undefined) {
+      console.warn('[@suspensive/react-query]', `Module ${moduleName} is not found`)
+      exit(1)
+    }
+
+    const functions = Object.keys(moduleExports)
+      .filter((key) => typeof moduleExports[key] === 'function')
+      .map((fn) => (fn[0] === fn[0].toUpperCase() ? `<${fn}/>` : fn))
+      .reverse()
+
+    results.push(...functions)
+  }
+
+  return results
 }
--- a/packages/react-query/src/scripts/utils/table.ts
+++ b/packages/react-query/src/scripts/utils/table.ts
@@ -1,32 +1,17 @@
 import Table from 'cli-table3'
-import { getPackageJson, getSuspensiveReactQueryPackageJson, getTanStackReactQueryPackageJson } from './package'
-
-export const VERSION5_APIS = [
-  '<SuspenseQuery/>',
-  '<SuspenseQueries/>',
-  '<SuspenseInfiniteQuery/>',
-  '<Mutation/>',
-  '<QueryErrorBoundary/>',
-]
-
-export const VERSION4_APIS = [
-  'useSuspenseQuery',
-  'useSuspenseQueries',
-  'useSuspenseInfiniteQuery',
-  'queryOptions',
-  'infiniteQueryOptions',
-  '<SuspenseQuery/>',
-  '<SuspenseQueries/>',
-  '<SuspenseInfiniteQuery/>',
-  '<Mutation/>',
-  '<QueryErrorBoundary/>',
-]
+import {
+  getPackageJson,
+  getSuspensiveReactQueryPackageJson,
+  getTanStackReactQueryPackageJson,
+  getTargetSuspensiveReactQueryAPIs,
+} from './package'
 
 export function getStatusTable(currentTargetVersion: string) {
   const packageJson = getPackageJson()
   const tanStackReactQueryPackageJson = getTanStackReactQueryPackageJson()
   const tanStackReactQueryMajorVersion = tanStackReactQueryPackageJson.version.split('.')[0]
   const targetSuspensiveReactQueryPackageJson = getSuspensiveReactQueryPackageJson(tanStackReactQueryMajorVersion)
+  const targetSuspensiveReactQueryAPIs = getTargetSuspensiveReactQueryAPIs()
 
   const table = new Table({
     head: [packageJson.name, 'result', 'status', 'advice'],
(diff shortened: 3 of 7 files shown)


packages/react-query/src/scripts/utils/package.ts
```diff
@@ -49,3 +51,41 @@ export function getSuspensiveReactQueryPackageJson(targetVersion: string) {
...
+  return fs.readFileSync(path.join(basePath, 'index.js'), 'utf-8') || ''
+}
+
+export function getTargetSuspensiveReactQueryVersion(): string {
+  const indexFileContent = getIndexFileContent()
+
+  return (RegExp(/@suspensive\/react-query-(\d+)/).exec(indexFileContent) || [])[1] || ''
+}
+
+export function getTargetSuspensiveReactQueryAPIs() {
```
gwansikk, line 67: Read the built JavaScript (index.js) file to check the current target @suspensive/react-query, parse it, and create a list of the exported APIs to display.

packages/react-query/src/scripts/utils/package.ts
```diff
@@ -49,3 +51,41 @@ export function getSuspensiveReactQueryPackageJson(targetVersion: string) {
...
+      console.warn('[@suspensive/react-query]', `Module ${moduleName} is not found`)
+      exit(1)
+    }
+
+    const functions = Object.keys(moduleExports)
+      .filter((key) => typeof moduleExports[key] === 'function')
+      .map((fn) => (fn[0] === fn[0].toUpperCase() ? `<${fn}/>` : fn))
+      .reverse()
+
+    results.push(...functions)
```
manudeli, line 87: If the name of a function needs to be provided as `<ComponentName/>` by distinguishing whether it is a function component based on its uppercase or not, let's remove the `<ComponentName/>` convention for now. This is because functions that start with an uppercase letter are not always guaranteed to be function components.

If so, I think you can change it like the suggestion code. I believe that ensuring our intent in code is a much more important priority than convention.

```suggestion
    results.push(...Object.keys(moduleExports).reverse())
```

packages/react-query/src/scripts/utils/table.ts
```diff
@@ -1,32 +1,17 @@
...
-  'useSuspenseQueries',
-  'useSuspenseInfiniteQuery',
-  'queryOptions',
-  'infiniteQueryOptions',
-  '<SuspenseQuery/>',
-  '<SuspenseQueries/>',
-  '<SuspenseInfiniteQuery/>',
-  '<Mutation/>',
-  '<QueryErrorBoundary/>',
-]
```
manudeli, line 23: Flexible! 🚀🚀🚀

packages/react-query/src/scripts/utils/package.ts
```diff
@@ -49,3 +51,41 @@ export function getSuspensiveReactQueryPackageJson(targetVersion: string) {
 
   return packageJson
 }
+
+export function getIndexFileContent() {
+  const basePath = path.resolve(__dirname, '../../dist').replace(/src/, '')
+
+  return fs.readFileSync(path.join(basePath, 'index.js'), 'utf-8') || ''
+}
```
manudeli, line 59: (Priority low) In my opinion we could __dirname as parameter to fix start position regardless of getIndexFileContent's executing position.

```suggestion
export function getIndexFileContent(dirname: string) {
  const basePath = path.resolve(dirname, 'dist').replace(/src/, '')

  return fs.readFileSync(path.join(basePath, 'index.js'), 'utf-8') || ''
}
```

packages/react-query/src/scripts/utils/loadModule.ts
```diff
@@ -1,7 +1,7 @@
 export function loadModule<T>(name: string): T | undefined {
```
manudeli, line 1: How about use typescript discriminated union

```suggestion
export function loadModule<T>(name: string): { exports: T, isSuccess: true } | { exports: undefined, isSuccess: false } {
```

### AS IS

```ts
const modules = indexFileContent.matchAll(/export \* from ['"](.+?)['"]/g)

for (const [, moduleName] of modules) {
  const moduleExports = loadModule<Record<string, unknown>>(moduleName)

  if (moduleExports === undefined) {
    console.warn('[@suspensive/react-query]', `Module ${moduleName} is not found`)
    exit(1)
  }

  const { ...interfaces } = moduleExports
}
```

### TO BE

```ts
const modules = indexFileContent.matchAll(/export \* from ['"](.+?)['"]/g)

for (const [, moduleName] of modules) {
  const module = loadModule<Record<string, unknown>>(moduleName)

  if (module.isSuccess) { // in my opinion use can success easily
    const { ...interfaces } = module.exports
  }

  console.warn('[@suspensive/react-query]', `Module ${moduleName} is not found`)
  exit(1)
}
```

packages/react-query/src/scripts/utils/package.ts
gwansikk: > If the name of a function needs to be provided as `<ComponentName/>` by distinguishing whether it is a function component based on its uppercase or not, let's remove the `<ComponentName/>` convention for now. This is because functions that start with an uppercase letter are not always guaranteed to be function components.

I agree. Starting with a capital letter does not guarantee that it is a `Components`.

manudeli:
@gwansikk Resolve conflict please

packages/react-query/src/scripts/utils/loadModule.ts
gwansikk: Great! This will improve handling to make it easier and clearer.

packages/react-query/src/scripts/utils/package.ts
```diff
@@ -1,51 +1,85 @@
...
+export function getTargetSuspensiveReactQueryVersion(): string {
+  const indexFileContent = getIndexFileContent(__dirname, '../../')
+
+  return (RegExp(/@suspensive\/react-query-(\d+)/).exec(indexFileContent) || [])[1] || ''
+}
+
+export function getTargetSuspensiveReactQueryAPIs(): string[] {
+  const indexFileContent = getIndexFileContent(__dirname, '../../')
+
+  const modules = indexFileContent.matchAll(/export \* from ['"](@suspensive\/react-query-\d+)['"]/g)
```
gwansikk, line 70: I have changed it to extract APIs only from suspensive/react-query. This is to prevent the re-export of TanStack Query v5's APIs in the future. (I believe the re-export of TanStack Query should not be displayed in the status table.)

manudeli: In my opinion, re-exported apis of @suspensive/react-query should be displayed in status table. because actually they are exported by @suspensive/react-query too anyway.

I want to check why you believe the re-export of TanStack Query should not be displayed in the status table.

gwansikk: I think the re-export of TanStack Query is not a feature provided by Suspensive and could cause confusion for users.

- The re-export of TanStack Query is a tool (hint) to help with migration from v4 to v5. (It cannot be used to leverage Suspensive.)
- If displayed in the Status Table, users might mistakenly use it as a Suspensive feature.

manudeli: > I think the re-export of TanStack Query is not a feature provided by Suspensive and could cause confusion for users.

In my opinion, Re-exporting of TanStack Query could be "feature" anyway provided by Suspensive's intent too.

> - If displayed in the Status Table, users might mistakenly use it as a Suspensive feature.

In my opinion, using re-exported apis is not mistake. because anyway we exports them as public apis. then, how about mark it as "re-exporting"

## Results expected

### srq status with @tanstack/react-query@4

|@suspensive/react-query|result|
|------|---|
|Available interfaces from @suspensive/react-query-4|useSuspenseQuery, useSuspenseQueries, useSuspenseInfiniteQuery, queryOptions, SuspenseQuery, SuspenseQueries, SuspenseInfiniteQuery, Mutation|
|Available interfaces from @tanstack/react-query@4||

### srq status with @tanstack/react-query@5

|@suspensive/react-query|result|
|------|---|
|Available interfaces from @suspensive/react-query-5|SuspenseQuery, SuspenseQueries, SuspenseInfiniteQuery, Mutation|
|Available interfaces from @tanstack/react-query@5|useSuspenseQuery, useSuspenseQueries, useSuspenseInfiniteQuery, queryOptions|

gwansikk: Cool! I will implement it using that approach!

gwansikk: @manudeli Perhaps this approach might make it more complicated for users. As you mentioned, re-export can be considered a "feature" (a feature that helps with migration). I would like to follow your opinion on how to display these APIs. What do you suggest we should do?

image

manudeli: How about this?

|@suspensive/react-query|result|status|description|
|------|---|---|---|
|version|2.9.0|🟢||
|export|@suspensive/react-query-5@2.9.0|🟢|You can use SuspenseQuery, SuspenseQueries, SuspenseInfiniteQuery, Mutation|
|peerDependency|@tanstack/react-query@5.29.1|🟢|You can use useSuspenseQuery, useSuspenseQueries, useSuspenseInfiniteQuery, queryOptions|

|@suspensive/react-query|result|status|description|
|------|---|---|---|
|version|2.9.0|🟢||
|export|@suspensive/react-query-4@2.9.0|🔴|You should `npx srq switch 5` to fix this error|
|peerDependency|@tanstack/react-query@5.29.1|🟢|You can use useSuspenseQuery, useSuspenseQueries, useSuspenseInfiniteQuery, queryOptions|

gwansikk:
@manudeli Cool! It seems clearer than before. Thank you.

image

manudeli (approved):
🚀 Dynamic table! cool

Issue #1076, opened 2024-07-13
[Feature]: remove Suspensive -> rename as DefaultProps


manudeli:
### Package Scope

@suspensive/react

### AS IS

```tsx
import { Suspensive, SuspensiveProvider } from '@suspensive/react'
 
const suspensive = new Suspensive({
  defaultProps: {
    delay: {
      ms: 1200,
    },
    suspense: {
      fallback: <Spinner />,
      clientOnly: false,
    },
  },
})
 
function App() {
  return <SuspensiveProvider value={suspensive}>...</SuspensiveProvider>
}

```

### TO BE

```tsx
import { DefaultProps, DefaultPropsProvider, Cache, CacheProvider} from '@suspensive/react'
 
const defaultProps = new DefaultProps({
  Delay: {
    ms: 1200,
  },
  Suspense: {
    fallback: <Spinner />,
    clientOnly: false,
  },
})

const cache = new Cache({
  staleTime: 2000
})
 
function App() {
  return (
    <CacheProvider cache={cache}>
      <DefaultPropsProvider defaultProps={defaultProps}>
        ...
      </DefaultPropsProvider>
    </CacheProvider>
  )
}

```

manudeli:
I'll support DefaultPropsProvider, SuspensiveProvider samely, and mark SuspensiveProvider as deprecated jsdoc. then I'll remove it at Suspensive v3

PR #1123, opened 2024-07-24
fix(react): deprecate DevMode

manudeli:
# Overview

I think this change might be perceived as a BREAKING CHANGE. However, this feature is already officially provided by react devtools, and I think it is better to guide it to an officially supported tool, and since it can reduce the bundle size by over half, Increasing the size this much is not what we intended and I think this could also be seen as a bug. I plan to remove it without upgrading the major version. 

I'm trying to get rid of it quickly so I can focus more on adding other features.

## Official alternative

Jul-24-2024 15-37-35

### AS-IS

image

### TO-BE

image

## PR Checklist


manudeli:
After this pr, I'll work related with https://github.com/toss/suspensive/issues/1076#issuecomment-2247008517

gwansikk:
If there are users who are not using react-devtools but are using the dev mode, I worry that this sudden change might be disruptive.

I think it would be appropriate to treat this as a BREAKING CHANGE. Alternatively, we could remove the feature but leave an implementation that shows a message indicating that the component cannot be used, keeping it as ‘@deprecated’ before eventually removing it.

manudeli:
> If there are users who are not using react-devtools but are using the dev mode, I worry that this sudden change might be disruptive.
> 
> in my opinion i think it would be appropriate to treat this as a BREAKING CHANGE. Alternatively, we could remove the feature but leave an implementation that shows a message indicating that the component cannot be used, keeping it as `@deprecated` before eventually removing it.

Cool, I'll remove feature but remain interface only typely with deprecated jsdoc but not work. and then I'll do removing type at v3

manudeli:
After [91561a5](https://github.com/toss/suspensive/pull/1123/commits/91561a520bb1b0b890bb56d4fd4de5b034f2b4a8)

Size slightly improved but It's not big deal and Removing DevMove gradually is available too! Thanks cool idea @gwansikk I'll add you as co-author
image

packages/react/src/contexts/SuspensiveDevModeContext.tsx
```diff
@@ -1,488 +0,0 @@
...
-            >
-              <stop stopColor="white" />
-              <stop offset="1" stopColor="white" stopOpacity="0.4" />
-            </radialGradient>
-          </defs>
-        </svg>
-      </button>
-    </aside>
-  )
-})
```
manudeli, line 488: We can also remove react-dom, @types/react-dom and big size unnecessary svg

gwansikk (approved):

.changeset/shaggy-baboons-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@suspensive/react": patch
+---
+
+fix(react): remove unnecessary DevMode
```
manudeli, line 5:
```suggestion
fix(react): deprcate DevMode
```

manudeli:
@kangju2000 @bigsaigon333 Could you review it please?

configs/eslint-config-js/no-import.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
       'warn',
       {
         cspell: {
-          words: ['packlint', 'codecov', 'tsup', 'nextra'],
+          words: ['packlint', 'codecov', 'tsup', 'nextra', 'Sandpack'],
```
kangju2000, line 30: 👍👍

packages/react/src/ErrorBoundary.tsx
```diff
@@ -71,7 +69,14 @@ export type ErrorBoundaryProps = PropsWithDevMode<
...
   }>,
-  ErrorBoundaryDevModeProp
+  {
+    showFallback?:
+      | boolean
+      | {
+          errorMessage?: string
+          after?: number
+        }
+  }
```
kangju2000, line 79: Is showFallback also going to be deprecated? If so, it would be great to note that in the JSDoc.

packages/react/src/Suspense.tsx
```diff
@@ -1,9 +1,9 @@
 import { type SuspenseProps as ReactSuspenseProps, useContext } from 'react'
-import { SuspenseDefaultPropsContext, syncDevMode } from './contexts'
+import { SuspenseDefaultPropsContext } from './contexts'
 import type { PropsWithDevMode } from './utility-types'
 import { defineSuspense } from './utils'
 
-export interface SuspenseProps extends PropsWithDevMode<ReactSuspenseProps, SuspenseDevModeProp> {
+export interface SuspenseProps extends PropsWithDevMode<ReactSuspenseProps, { showFallback?: boolean }> {
```
kangju2000, line 6: It would be good to use JSDoc here as well, just like in the comment above.

packages/react/src/ErrorBoundary.tsx
manudeli: Thanks! I added you as co-author

kangju2000 (approved):

manudeli:
We can check how big size is reduced by this change

Screenshot_20240724_185441_Chrome

So much good commenter made by @kangju2000 You are king 👑

manudeli:
@kangju2000 I want to check size .next folder built by visualization project depend on all Suspensive packages

Could do make new pull request to check size change of real services depending Suspensive?

bigsaigon333 (approved):
Greeeeeeeeat!!!

I didn't realize that the official React DevTools could be used as an alternative to DevMode, but the GIF you provided above makes it so easy to understand 👍🏻👍🏻

manudeli:
Released npm size comparison

We size off 504kB, 61% (813kB -> 309kB)

### AS-IS (@suspensive/react@2.9.1)

image

### TO-BE (@suspensive/react@2.9.2)

image

Issue #1126, opened 2024-07-24
[Feature]: check size change of visualization project depending Suspensive


kangju2000:
### Package Scope

etc

### Description

https://github.com/toss/suspensive/pull/1123#issuecomment-2247457356

Add script to check size change of real services(visualization project) depending Suspensive.

kangju2000:
@manudeli 

Can I do this by adding to the pattern in the compressed-size-action script?

manudeli:
Yes, I thought that way to do this👍

PR #1127, opened 2024-07-24
chore: update compressed-size-action pattern to check size of visualization project

kangju2000:
# Overview

close https://github.com/toss/suspensive/issues/1126

Add script to check size change of real services(visualization project) depending Suspensive.

## PR Checklist

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Report build size
         uses: preactjs/compressed-size-action@v2
         with:
-          build-script: 'ci:size'
+          build-script: '--filter ./websites/visualization build'
           pattern: '{packages/**/dist/*.{js,cjs,mjs},websites/visualization/.next/static/chunks/**/*.js,websites/visualization/.next/server/pages/*.{html,json}}'
           exclude: '{**/*.map,**/node_modules/**}'
           strip-hash: "\\w+-(\\w{16}).js$"
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -39,9 +39,9 @@ jobs:
       - name: Report app size
         uses: preactjs/compressed-size-action@v2
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/compressed-size.yml
```diff
@@ -23,6 +23,8 @@ jobs:
       - name: Report build size
         uses: preactjs/compressed-size-action@v2
         with:
-          build-script: 'prepack'
-          pattern: 'packages/**/dist/*.{js,cjs,mjs}'
+          build-script: 'build'
+          pattern: '{packages/**/dist/*.{js,cjs,mjs},websites/visualization/.next/static/chunks/**/*.js,websites/visualization/.next/server/pages/*.{html,json}}'
```
manudeli, line 27: Do we have way to check size of .next directory not each files?

kangju2000: `compressed-size-action` checks the size of individual files, so it's probably not possible to show the overall size change of the `.next` folder.

## What I tried
- Run `compressed-size-action` as two separate scripts 
  - it didn't create two comments..



## Possible Solution
### Add a different actions to create a new comment for this
But I'm worried that this might lead to too many comments on each PR, making it hard to read.
Also, when an outsider looks at this comment, it might be difficult for them to immediately know whether we're bundling the `visualization website` or `suspensive.org`.


- [actions-next-bundle-analyzer](https://github.com/transferwise/actions-next-bundle-analyzer)

- implement it myself [reference blog](https://jeffchen.dev/posts/Measuring-Bundle-Sizes-With-Next-js-And-Github-Actions/)

What do you think I should do? If you have any good alternatives, please let me know 🙏

manudeli: > [actions-next-bundle-analyzer](https://github.com/transferwise/actions-next-bundle-analyzer)

Oh this could be good to check next built size...! could you do this for this repository?

manudeli (approved):
I'll try this in https://github.com/toss/suspensive/pull/1123 for now! If we have problem, Let's fix it at that time!

PR #1132, opened 2024-07-25
chore: increase the speed of the size action in CI

kangju2000:
# Overview

Improve CI speed through the prepack script.

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,8 +53,8 @@ jobs:
       - if: matrix.command == 'prepack'
         uses: preactjs/compressed-size-action@v2
         with:
-          install-script: 'echo "skipping install"'
-          build-script: 'echo "skipping build"'
+          install-script: 'noop "skip install"'
+          build-script: 'noop "skip build"'
           pattern: 'packages/**/dist/*.{js,cjs,mjs}'
           exclude: '{**/*.map,**/node_modules/**}'
           repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -61,3 +62,4 @@ jobs:
         with:
           working-directory: ./websites/visualization
           comment-strategy: 'always'
```
kangju2000, line 64: [`skip-insignificant`](https://github.com/transferwise/actions-next-bundle-analyzer?tab=readme-ov-file#usage) is skip commenting on PRs if there are no signficant changes in page sizes.

Do you think we should always show the bundle size of the visualization website, even if there's no change in size?

.github/workflows/ci.yml
```diff
@@ -61,3 +62,4 @@ jobs:
...
+          create-issue: false
```
kangju2000, line 65: I'll make sure no issue is created. https://github.com/toss/suspensive/issues/1128

manudeli: But if we need to diff default branch with Pull Request branch, don't we need to create issue?

.github/workflows/ci.yml
```diff
@@ -53,6 +53,7 @@ jobs:
       - if: matrix.command == 'prepack'
         uses: preactjs/compressed-size-action@v2
         with:
+          build-script: 'prepack'
```
manudeli, line 56: 👍

.github/workflows/ci.yml
kangju2000: The issue only records the current bundle size of the website, so creating an issue isn't essential for the action to run properly.

If you want the bundle size to always be updated through issues, we can keep generating the issues!

What do you think?

manudeli: > If you want the bundle size to always be updated through issues

I don't want to make issue too. Thanks!

manudeli (approved):

PR #1157, opened 2024-07-28
test(react): improve `ErrorBoundaryGroup` test code

jgjgill:
# Overview

Hi, I've been looking at the `ErrorBoundaryGroup` test code and trying to improve it.

- I changed `Throw.reset()` to be managed by `beforeEach`.
- `useErrorBoundaryGroup` related tests only have tests that fail. I add success case.

**Question**

[Related Code](https://github.com/toss/suspensive/blob/cbbb6ffe649cfd4442da5fcf55e92d0afdf0fb3c/packages/utils/src/test-utils/index.tsx#L4-L26)

The `current` value of `isNeedThrowGlobal` used by the `Throw` object doesn't seem to change.
There doesn't seem to be anything currently happening that would cause `isNeedThrowGlobal.current = true`.

스크린샷 2024-07-28 오후 6 54 04

So the absence of `Throw.reset()` in the current test doesn't seem to change the result of the test.
Was this code added for safety?

Thanks. 🙇‍♂️

## PR Checklist

### DIFF
--- a/packages/react/src/ErrorBoundaryGroup.spec.tsx
+++ b/packages/react/src/ErrorBoundaryGroup.spec.tsx
@@ -13,6 +13,8 @@ const innerErrorBoundaryCount = 3
 const resetButtonText = 'reset button'
 
 describe('<ErrorBoundaryGroup/>', () => {
+  beforeEach(() => Throw.reset())
+
   it('should reset all ErrorBoundaries in children', async () => {
     render(
       <ErrorBoundaryGroup>
@@ -31,7 +33,6 @@ describe('<ErrorBoundaryGroup/>', () => {
 
     expect(screen.getAllByText(TEXT).length).toBe(innerErrorBoundaryCount)
     await waitFor(() => expect(screen.getAllByText(ERROR_MESSAGE).length).toBe(innerErrorBoundaryCount))
-    Throw.reset()
 
     fireEvent.click(screen.getByRole('button', { name: resetButtonText }))
     expect(screen.getAllByText(TEXT).length).toBe(innerErrorBoundaryCount)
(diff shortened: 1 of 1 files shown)


packages/react/src/ErrorBoundaryGroup.spec.tsx
```diff
@@ -67,7 +67,21 @@ describe('<ErrorBoundaryGroup/>', () => {
...
+          {createElement(() => {
+            useErrorBoundaryGroup()
+            return <></>
+          })}
+          <ErrorBoundary fallback={ERROR_MESSAGE}>
+            <>{TEXT}</>
+          </ErrorBoundary>
+        </ErrorBoundaryGroup>
+      ).getByText(TEXT)
+    ).toBeInTheDocument()
```
manudeli, line 83: We forgot happy path at that time! Cool!

manudeli (approved):
Thanks! 👍

manudeli:
@kangju2000 https://github.com/toss/suspensive/actions/runs/10130818442/job/28012703192?pr=1157 check this action's error please

kangju2000:
> @kangju2000 https://github.com/toss/suspensive/actions/runs/10130818442/job/28012703192?pr=1157 check this action's error please
> 
> In my opinion, removing it is also good for me. We can just check only sizes of packages too


@manudeli 

It seems like a token-related issue since there is a difference in CI permissions between maintainers and contributors.

|maintainer|contributor|
|---|---|
|image|image|



<br />

Could you please adjust the workflow permissions in the settings? If it's already set up, I'll remove the action.

image
> https://github.com/orgs/community/discussions/60820#discussioncomment-6445895

manudeli:
@kangju2000 

image

Yeah, I setup like your suggestion to use changeset's action. First, Let's remove it and then reconsider it later. 🙇

PR #1160, opened 2024-07-28
feat(react): add DefaultProps & DefaultPropsProvider

manudeli:
related with https://github.com/toss/suspensive/issues/1076#issuecomment-2247008517

# Overview

Suspensive is used only to provide defaultProps.
I want to make it more intuitively.

So I deprecate Suspensive, SuspensiveProvider and add DefaultProps DefaultPropsProvider
I want to remove Suspensive, SuspensiveProvider in v3

## PR Checklist


.changeset/odd-hornets-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@suspensive/react": minor
+---
+
+feat(react): add DefaultProps & DefaultPropsProvider
```
manudeli, line 5:
```suggestion
feat(react): add DefaultProps & DefaultPropsProvider and deprecate Suspensive & SuspensiveProvider
```

websites/visualization/src/app/providers.tsx
```diff
@@ -1,47 +1,25 @@
...
-          },
-          suspense: {
-            fallback: <Spinner />,
-          },
-        },
-      })
-  )[0]
+  const cacheStore = useState(() => new CacheStore())[0]
+  const queryClient = useState(() => new QueryClient({ defaultOptions: { queries: { retry: 0 } } }))[0]
+  const defaultProps = useState(() => new DefaultProps({ delay: { ms: 1200 }, suspense: { fallback: <Spinner /> } }))[0]
```
manudeli, line 13: Do we need to use uppercase?

```suggestion
  const defaultProps = useState(() => new DefaultProps({ Delay: { ms: 1200 }, Suspense: { fallback: <Spinner /> } }))[0]
```

SEOKKAMONI: I think they should be in lowercase. Users are generally more accustomed to seeing uppercase when it refers to a class or component. Considering this, since both "suspense" and "delay" are not directly classes or components, they should be in lowercase to be more familiar and approachable for users.

manudeli: But as you know, why I raise this naming issue is that we use uppercases for naming of each components(Suspense, Delay). If we are using uppercase for component naming, Isn't it proper that we use uppercase for target component field naming of defaultProps too?

Isn't it good? I thought it as consistency problem

manudeli: What I mean is that I don't think there is a reason to use lowercase letters, rather than that it has to be capitalized. I think it would be better to use the same capitalization for the names of components, from an intuitive perspective.

I'm having a hard time deciding.

SEOKKAMONI: Ah, I understand now. You meant that ms and fallback should be specified as default values for each Suspense and Delay component, so they should be in uppercase. If that’s the intention, I think using uppercase is more intuitive. I believe it’s a good idea!

SEOKKAMONI: @gwansikk @kangju2000 @bigsaigon333 I’m curious about what others think as well. What are your thoughts?

gwansikk: I think using lowercase is better. This is because it's not rendering an actual component but passing an option. Lowercase is commonly used for options, and using uppercase might make users hesitate and think they need to provide a value in the form of a component. (Of course, the type is specified, but I think it's better to make it clear that it's an option.)

kangju2000: Before that, I'm curious why you want to change it to a more intuitive name. `DefaultProps` seems like a generic component without context. Could you explain in detail why you made this change? @manudeli

SEOKKAMONI: In my opinion, since the interface is named DefaultProps, I think users will naturally understand it as specifying the default values for a particular component. If it is perceived as I think, then using capitalized names, such as Delay or Suspense, to specify the options would align better with the naming of the interface.

manudeli: @kangju2000, @gwansikk 

- https://github.com/toss/suspensive/pull/1160#discussion_r1695109541
- https://github.com/toss/suspensive/pull/1160#discussion_r1695116805

I want to check whether you guys read these comments. I don't think there is a absolute right choice, but I think we should check each other's thoughts and decide whether to capitalize or lowercase. Let me know whether you understood intention when using uppercase.

## Wrap up

### Uppercase
```tsx
const defaultProps = new DefaultProps({
  Suspense: { // Uppercase: Same with target component
    fallback: <Spinner />
  },
  ...
})

const Example = () => {
  return (
    <DefaultPropsProvider defaultProps={defaultProps}>
      <Suspense /> // default fallback should be <Spinner />
    </DefaultPropsProvider>
  )
}
```

### Lowercase
```tsx
const defaultProps = new DefaultProps({
  suspense: { // Lowercase: Different with target component
    fallback: <Spinner />
  },
  ...
})

const Example = () => {
  return (
    <DefaultPropsProvider defaultProps={defaultProps}>
      <Suspense /> // default fallback should be <Spinner />
    </DefaultPropsProvider>
  )
}
```

gwansikk: Oh I agree, I realized it after seeing your example, **uppercase** would be more consistent and natural in that case.

Injected as `DefaultProps.TargetComponent.props(options)` is more natural

manudeli: After @kangju2000 approving renaming, I'll capitalize as below

```tsx
const defaultProps = new DefaultProps({
  Suspense: { // Uppercase: Same with target component
    fallback: <Spinner />
  },
  ...
})

const Example = () => {
  return (
    <DefaultPropsProvider defaultProps={defaultProps}>
      <Suspense /> // default fallback should be <Spinner />
    </DefaultPropsProvider>
  )
}

```

kangju2000: @manudeli 
I agree as well. It seems better to keep the keys in uppercase to indicate that they represent components.

manudeli:
@gwansikk @kangju2000 @bigsaigon333 @SEOKKAMONI I reflected our reviews
Review this pull request to add new api(DefaultProps, DefaultPropsProvider) please

docs/suspensive.org/src/pages/docs/react/DefaultPropsProvider.en.mdx
```diff
@@ -0,0 +1,21 @@
+# DefaultPropsProvider
+
+This component controls the settings of the components provided by Suspensive (Suspense, Delay, etc.) at once.
+
+```jsx /DefaultPropsProvider/ /DefaultProps/
+import { DefaultProps, DefaultPropsProvider } from '@suspensive/react'
+
+const defaultProps = new DefaultProps({
+  delay: {
```
gwansikk, line 9:
```suggestion
  Delay: {
```
It seems that the documentation update has been missed.

docs/suspensive.org/src/pages/docs/react/DefaultPropsProvider.en.mdx
```diff
@@ -0,0 +1,21 @@
...
+    ms: 1200,
+  },
+  suspense: {
```
gwansikk, line 12:
```suggestion
  suspense: {
```

manudeli:
@gwansikk You have 👀🦅

gwansikk (approved):
ready for suspensive! 🚀🚀

Issue #1071, opened 2024-07-11
[Feature]: Support lazy load `<InView />`


manudeli:
### Package Scope

etc

### Description

@suspensive/intersection to trigger fetch on view inside of window

### Possible Solution

```tsx
const posts = useSuspenseQuery(postsQueryOptions(...))

return posts.map((post) => (
    <Intersection fallback={<Skeleton />}>
      <Suspense fallback={<Skeleton />}>
        <SuspenseQuery {postQueryOptions(post.id)}>
          {(post) => <Post {...post} />}
        </SuspenseQuery>
      </Suspense>
    </Intersection>
  )
)
```

manudeli:
```
const posts = useSuspenseQuery(postsQueryOptions(...))

return posts.map((post) => (
  <Suspense intersection fallback={({ intersectionRef }) => <Skeleton ref={intersectionRef} />}>
    <SuspenseQuery {postQueryOptions(post.id)}>
      {(post) => <Post {...post} />}
    </SuspenseQuery>
  </Suspense>
  )
)
```

@gwansikk @kangju2000 @bigsaigon333
This could be new intersection prop of Suspense to lazy load??

gwansikk:
> ```tsx
> 
> const posts = useSuspenseQuery(postsQueryOptions(...))
> 
> 
> 
> return posts.map((post) => (
> 
>   <Suspense intersection fallback={({ intersectionRef }) => <Skeleton ref={intersectionRef} />}>
> 
>     <SuspenseQuery {postQueryOptions(post.id)}>
> 
>       {(post) => <Post {...post} />}
> 
>     </SuspenseQuery>
> 
>   </Suspense>
> 
>   )
> 
> )
> 
> ```
> 
> 
> 
> @gwansikk @kangju2000 @bigsaigon333
> 
> This could be new intersection prop of Suspense to lazy load??

Oh, that's a great idea! It seems useful for gradually loading the necessary data for users through lazy loading

minsoo-web:
However, I think Intersection Observe should have more options. 

For example,

[rootMargin](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/rootMargin)
[thresholds](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/thresholds)

I thought, "How do I set these options?"

minsoo-web:
```ts
const posts = useSuspenseQuery({
  ...otherOptions,
  observeRef: ref
})

[...]

PR #1184, opened 2024-08-03
feat(react-dom): add InView, useInView

manudeli:
related with #1071

# Overview

## PR Checklist


manudeli:
coauthors test

manudeli:
This is quite big change. Could I request this Pull Request you guys? @gwansikk @kangju2000 @bigsaigon333 @SEOKKAMONI

packages/react-dom/src/InView.tsx
```diff
@@ -0,0 +1,141 @@
...
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      delay,
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      initialInView,
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      fallbackInView,
+      ...props
+    } = this.props as PlainChildrenProps
+
+    return createElement(as || 'div', { ref: this.handleNode, ...props }, children)
```
SEOKKAMONI, line 139: I'm curious why destructuring assignment was used for unused properties.

packages/react-dom/src/InView.tsx
```diff
@@ -0,0 +1,141 @@
+import { Component, createElement } from 'react'
+import { observe } from './observe'
+import type { IntersectionObserverProps, PlainChildrenProps } from './types'
+
+function isPlainChildren(props: IntersectionObserverProps | PlainChildrenProps): props is PlainChildrenProps {
+  return typeof props.children !== 'function'
+}
```
SEOKKAMONI, line 7: The usage of isPlainChildren is singular and it's used as a negation, as shown below:

```tsx
 if (!isPlainChildren(this.props)) {
      // Store the current State, so we can pass it to the children in the next render update
      // There's no reason to update the state for plain children, since it's not used in the rendering.
      this.setState({ inView, entry })
    }
```

It's a minor suggestion, but how about using isFunctionChildren instead? I don't think there's a need to use a negation for isPlainChildren in this case.

packages/react-dom/src/setup.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { beforeAll, expect, test, vi } from 'vitest'
+import { mockAllIsIntersecting } from './test-utils'
+
+vi.hoisted(() => {
+  // Clear the `beforeEach` from global, so we can detect if this is a test env
+  // @ts-expect-error Property 'vi' does not exist on type 'typeof globalThis'.
+  global.vi = undefined
+})
+
+beforeAll(() => {})
```
SEOKKAMONI, line 10:
```suggestion
```

It seems like an unnecessary element.

SEOKKAMONI (approved):
It's a great interface!!

packages/react-dom/src/InView.tsx
manudeli: Oh! cool! I removed isPlainChildren. 

```diff
  handleChange = (inView: boolean, entry: IntersectionObserverEntry) => {
    if (inView && this.props.triggerOnce) {
      // If `triggerOnce` is true, we should stop observing the element.
      this.unobserve()
    }
-    if (!isPlainChildren(this.props)) {
+   if (this.props.children === 'function') {
      // Store the current State, so we can pass it to the children in the next render update
      // There's no reason to update the state for plain children, since it's not used in the rendering.
      this.setState({ inView, entry })
    }
    if (this.props.onChange) {
      // If the user is actively listening for onChange, always trigger it
      this.props.onChange(inView, entry)
    }
  }
```

packages/react-dom/src/InView.tsx
manudeli: I removed PlainChildren now! I think supporting children as only render function is better than supporting 2ways both. And InView component is simpler than before too.

gwansikk (approved):
cool !! 👍👍👍

PR #1197, opened 2024-08-04
feat(cache): add `Subscribable`

SEOKKAMONI:
# Overview
- add `Subscribable`

## PR Checklist


packages/cache/src/models/Subscribable.ts
```diff
@@ -0,0 +1,35 @@
+import type { CacheKey, CacheOptions, Sync } from '../types'
+import { hashCacheKey } from '../utils'
+
+export class Subscribable {
```
SEOKKAMONI, line 4: I chose to name the class `Subscribable` because it provides subscription functionality and manages subscribers, including synchronization. I believe the focus should be on managing subscribers rather than on synchronization itself, as synchronization is a part of subscriber management. Therefore, I decided on the name `Subscribable`, which better reflects this aspect from an external perspective.

packages/cache/src/models/Subscribable.ts
```diff
@@ -0,0 +1,37 @@
...
+  protected syncsMap: Map<ReturnType<typeof hashCacheKey>, Array<Sync>>
+
+  constructor() {
+    this.syncsMap = new Map()
+    this.subscribe = this.subscribe.bind(this)
```
ojj1123, line 9:
```ts
    this.subscribe = this.subscribe.bind(this)
```

[NIT]
Is it necessary to bind `this`?

SEOKKAMONI: For example, when passing a method as a parameter to another function, the reference to this can change, causing this to point to something other than what was originally intended. In such cases, explicitly binding this ensures that it refers to the expected object.

## When I didn't bind it
스크린샷 2024-08-15 오후 10 16 11

## When you bind it
스크린샷 2024-08-15 오후 10 16 19

SEOKKAMONI: To elaborate further, the subscribe method can be passed as a callback function or assigned to a variable, which may change the reference. For this reason, we use binding to explicitly specify what this refers to.

SEOKKAMONI: I hope that answer was helpful :)

manudeli:
This is comment to get co-author string

manudeli:
Share your proximate plan on this pull request please. I want to check what you have issue to do this after that date

packages/cache/src/models/Subscribable.ts
ojj1123: Oh I got it. Thanks!

manudeli (approved):
Let's rename CacheStore to Cache

PR #1204, opened 2024-08-09
docs: change default version

sungh0lim:
# Overview

- The default is v2, so the issue appears to be v2 even in the `https://suspensive.org`.

## PR Checklist


manudeli (approved):
Thanks! 👍

PR #1228, opened 2024-08-21
feat(lint): introducing eslint flat config

### LINKED ISSUE #1213: [Feature]: eslint flat config
### Package Scope

etc

### Description

eslint flat config

### Possible Solution

_No response_

### etc.

_No response_

vr-varad:
close #1213 

# Overview

## PR Checklist


vr-varad:
@manudeli I've implemented the ESLint flat config for eslint-config-ts, and it’s working perfectly. However, I changed the module type to "module" to support ES module imports. Should I revert to CommonJS modules instead? If so, this would require changing the module type for the entire package. What are your thoughts on this?

manudeli:
> @manudeli I've implemented the ESLint flat config for eslint-config-ts, and it’s working perfectly. However, I changed the module type to "module" to support ES module imports. Should I revert back to CommonJS modules instead? As, this would require changing the module type for the entire package. What are your thoughts on this? Before proceeding further, I just need confirmation that this is the right direction.

As far as I understand, esm can import cjs,
and cjs cannot require esm. Then I thought it would be better for these eslint-related packages to be written in cjs. Because packages like docs/suspensive.org use Next.js, they should be written in cjs.

Nevertheless, if you change the eslint-related packages to esm, I wonder if you can try using the exports field in package.json. It seems that you can make a package that supports both esm and cjs with tsup. (All packages under packages/* support this way, supporting both esm and cjs.)

Personally, I think it would be best to use "type: module" and make it support both esm and cjs with tsup and exports. Can you try this?

Then, I think what you should do is use type: module as it is now, add tsup, and check if it works with the exports field.

vr-varad:
@manudeli, I've converted all the configurations to flat config. However, making them compatible with both ESM and CJS has proven to be a bit tricky 😅. I’m working on it and expect to have it completed by tomorrow.

vr-varad:
@manudeli I've tried various approaches to solve this issue, but I haven't been able to find a working solution. Could you please help me out with this? Your expertise would be greatly appreciated!

vr-varad:
@manudeli any further follow-up?

manudeli:
> @manudeli any further follow-up?

Sorry, I haven't seen this PR lately as I've been busy with other things at the company. I'm trying to remove @suspensive/eslint-config-js, @suspensive/eslint-config-ts right now. Please wait a moment and I'll mention it again.

manudeli:
@vr-varad @gwansikk Could you guys check my update please? I updated eslint config based on vr-varad's Pull Request from eslint legacy config to eslint flat config

vr-varad:
Based on the code, it looks like you removed the `eslint-config-ts` and `eslint-config-js` folders and created a new `src` folder within `eslint-config`. Inside this `src` folder, you've added an `index.ts` file that consolidates the configurations from both the previous TypeScript and JavaScript ESLint configurations. Is that correct?

manudeli:
> Based on the code, it looks like you removed the `eslint-config-ts` and `eslint-config-js` folders and created a new `src` folder within `eslint-config`. Inside this `src` folder, you've added an `index.ts` file that consolidates the configurations from both the previous TypeScript and JavaScript ESLint configurations. Is that correct?

Yes, It is much simpler, supports cjs and esm, and also supports typescript. It was possible thanks to you gave me the basic idea! @vr-varad You're king 👍

manudeli:
> > @vr-varad @gwansikk Let me know if you okay too after you check please
> 
> LGTM! after merging this PR, I think we should also consider improving cases like this one
> 
> ```ts
> // packages/react/src/ErrorBoundary.tsx
> 
> // TODO: remove this line
> // eslint-disable-next-line @typescript-eslint/unbound-method
> onError,
> ```

Yeah! Thanks to @vr-varad, we can go to eslint@9 too! Let's improve after merging this pull request. If you okay approve this pr please

manudeli (approved):

manudeli (approved):
King of eslint @vr-varad, Thanks! 🙇

vr-varad:
And will continue to contribute to toss!!!!🙌🙌

gwansikk (approved):
Thank you for contribution! 👍👍👍 @vr-varad

PR #1229, opened 2024-08-22
chore(pnpm): use catalog

manudeli:
https://pnpm.io/catalogs


gwansikk (approved):
I agree with using the catalog protocol! The configuration has been set up correctly.

Optionally, what do you think about using [named-catalogs](https://pnpm.io/catalogs#named-catalogs)?

With React 19 coming soon, it might be challenging for all of our packages to support React 19 simultaneously. So, perhaps we could use named-catalogs to differentiate between React versions in advance? (This is just my personal opinion. I agree with this PR and will approve it as everything looks good!)

Example:
```yaml
catalogs:
  react18:
    react: ^18
    react-dom: ^18
    @types/react: ^18
# React 19 is just an example. I think it would be great to set it up like this for migration when ready-packages are compatible in the future.
  react19:
    react: ^19
    react-dom: ^19
    @types/react: ^19
```

manudeli:
> I agree with using the catalog protocol! The configuration has been set up correctly.
> 
> Optionally, what do you think about using [named-catalogs](https://pnpm.io/catalogs#named-catalogs)?
> 
> With React 19 coming soon, it might be challenging for all of our packages to support React 19 simultaneously. So, perhaps we could use named-catalogs to differentiate between React versions in advance? (This is just my personal opinion. I agree with this PR and will approve it as everything looks good!)
> 
> Example:
> 
> ```yaml
> catalogs:
>   react17:
>     react: ^17
>     react-dom: ^17
>     @types/react: ^17
> # React 19 is just an example. I think it would be great to set it up like this for migration when ready-packages are compatible in the future.
>   react19:
>     react: ^19
>     react-dom: ^19
>     @types/react: ^19
> ```

Oh..! Could you do it after merging this Pull Request? in my opinion, React Query version could be matter for your comment's point!

gwansikk:
> Oh..! Could you do it after merging this Pull Request? in my opinion, React Query version could be matter for your comment's point!

Sure, I'll try it after this PR is merged

Issue #1267, opened 2024-09-10
[Feature]: useQueryClient


gwansikk:
### Package Scope

@suspensive/react-query

### Description

ref: https://tkdodo.eu/blog/react-query-fa-qs#why-should-i-usequeryclient

> It is sometimes necessary to create the queryClient inside the App component (as shown above). One example is when using server side rendering, because you want to avoid having multiple users share the same client.

```tsx
<QueryClient>
  {(queryClient) => (
    <button
      onClick={() => queryClient.invalidateQueries({
        queryKey: ['items']
      })}
    >
      invalidate items
    </button>
  )}
</QueryClient>
```

gwansikk:
related: https://github.com/toss/suspensive/issues/1251

PR #1269, opened 2024-09-11
feat(react-query): add `<QueryClient>`

gwansikk:
# Overview

close: #1267 

## PR Checklist


manudeli:
I'm a little worried about the name QueryClient because it's an interface in @tanstack/query-core. I'd like to propose an alternative name: QueryClientConsumer. I want to know your opinion.

packages/react-query-5/src/QueryClientConsumer.tsx
```diff
@@ -0,0 +1,12 @@
+import { type QueryClient, useQueryClient } from '@tanstack/react-query'
+import type { ReactNode } from 'react'
+
+export function QueryClientConsumer({
```
manudeli, line 4: In my opinion, This need experimental jsdoc yet. Could you do that please?

packages/react-query-4/src/QueryClientConsumer.tsx
```diff
@@ -0,0 +1,12 @@
+import { type QueryClient, useQueryClient } from '@tanstack/react-query'
+import type { ReactNode } from 'react'
+
+export function QueryClientConsumer({
```
manudeli, line 4: In my opinion, This need experimental jsdoc yet. Could you do that please?

packages/react-query-4/src/QueryClientConsumer.tsx
```diff
@@ -0,0 +1,12 @@
...
+  children,
+  context,
+}: Readonly<{
+  children: (queryClient: QueryClient) => ReactNode
+  context?: React.Context<QueryClient | undefined>
+}>) {
```
manudeli, line 10:
```suggestion
}: {
  children: (queryClient: QueryClient) => ReactNode
  context?: React.Context<QueryClient | undefined>
}) {
```

Do we need Readonly? I want to know your intention

gwansikk: My intention was to prevent the defined `queryClient` or `context` from being reassigned due to user error. TanStack Query has a high dependency on `queryClient`, so I wanted to reduce mistakes by preventing changes.

Upon reflection, I believe this was a hasty approach, so I have reverted the changes to their original state.

.changeset/flat-games-unite.md
```diff
@@ -0,0 +1,7 @@
+---
+'@suspensive/react-query-4': minor
+'@suspensive/react-query-5': minor
+'@suspensive/react-query': minor
+---
+
+feat(react-query): add `<QueryClientConsumer>`
```
manudeli, line 7:
```suggestion
feat(react-query): add `<QueryClientConsumer/>`
```

PR #1294, opened 2024-10-06
feat(react-query): add `<HydrationBoundary/>`

gwansikk:
# Overview

## PR Checklist


manudeli:
I'm curious why we need to add `<HydrationBoundary/>` than using just [`<Hydration/>`](https://tanstack.com/query/v4/docs/framework/react/reference/hydration#hydrate-1) of @tanstack/react-query

PR #1302, opened 2024-10-11
refactor(react): use direct `undefined` comparison for fallback in `<Delay/>` and `<Suspense/>` components

dayongkr:
# Overview

Previously, `typeof` was used to check if `fallback` was `undefined`, but now it directly compares against `undefined` for improved efficiency. Additionally, added missing test cases for the `fallback` functionality.

## PR Checklist

## Additional

I was impressed by how `defaultProps` can be conveniently managed using `useContext`! I learned a lot thanks to this.


manudeli (approved):
Thanks! 🚀

packages/react/src/Delay.tsx
```diff
@@ -9,18 +9,16 @@ export interface DelayProps extends PropsWithChildren {
...
+    SuspensiveError.assert(props.ms >= 0, Message_Delay_ms_prop_should_be_greater_than_or_equal_to_0)
   }
   const defaultProps = useContext(DelayDefaultPropsContext)
   const ms = props.ms ?? defaultProps.ms ?? 0
 
   const [isDelaying, setIsDelaying] = useState(ms > 0)
   useTimeout(() => setIsDelaying(false), ms)
 
-  const fallback = typeof props.fallback === 'undefined' ? defaultProps.fallback : props.fallback
+  const fallback = props.fallback === undefined ? defaultProps.fallback : props.fallback
```
manudeli, line 21: That makes sense

dayongkr:
> > I was impressed by how defaultProps can be conveniently managed using useContext! I learned a lot thanks to this.
> 
> In projects where accuracy is more important than aesthetics, such as admin projects, it is often more important to just fill the fallback of `<Suspense/>` by whatever possible, so we provide this DefaultProps feature. I usually fill default fallback of `<Suspense/>` by just common `<Spinner/>` and override it by each fallback prop of `<Suspense/>`. @dayongkr I am also glad to see you in suspensive too
> 
> es-toolkit Contributions are good, but please come here often too kk

I will contribute more frequently! Thank you for the warm welcome~

PR #1312, opened 2024-10-13
feat(react): function children of Delay

### LINKED ISSUE #1300: [Feature]: Delay's function children
### Package Scope

@suspensive/react

### Description

```jsx
const Example () => {
  return (
    // Prohibit fallback by typescript function overriding
    <Delay ms={2000}>
      {({ isDelayed }) => (
         isDelayed ? <></> : null
       )}
    </Delay>
  )
}
```

### Possible Solution

_No response_

### etc.

_No response_

manudeli:
close #1300 

# Overview

### Problem

```tsx
const Example = () => {
  const useState(false)

  return (
    <Delay ms={1000}>
      {/* just expose Modal ui after 1000ms. if we need to open with delay, we can't support it */}
      <Modal isOpen={!isDelayed} /> 
    </Delay>
  )
}
```

### Solution

```tsx
const Example = () => {
  const useState(false)

  return (
    <Delay ms={1000}>
      {({ isDelayed }) => (
        // We can support from now!
        <Modal isOpen={!isDelayed} />
      )}
    </Delay>
  )
}
```

But, I'm not confident for this feature. so I marked this feature as experimental feature

## PR Checklist


gwansikk (approved):
cool 👍👍👍

packages/react/src/Delay.tsx
```diff
@@ -15,11 +25,20 @@ export const Delay = (props: DelayProps) => {
...
+  }
 
-  const fallback = props.fallback === undefined ? defaultProps.fallback : props.fallback
-  return <>{isDelaying ? fallback : props.children}</>
+  if (isDelayed) {
+    return <>{props.children}</>
+  }
+  if (props.fallback === undefined) {
+    return <>{defaultProps.fallback}</>
+  }
```
kangju2000, line 40: Is it correct behavior that when props.fallback is null, the fallback from defaultProps doesn't activate?

manudeli: That's right! I added test code to prove this as spec in [8960583](https://github.com/toss/suspensive/pull/1312/commits/896058394720a8bd770d1c2bd358db1ccf83b351) and I added you as co-author

packages/react/src/Delay.tsx
```diff
@@ -1,12 +1,22 @@
...
-  fallback?: ReactNode
-}
+export type DelayProps =
+  | {
+      ms?: number
+      fallback?: never
+      /**
+       * @experimental This is experimental feature.
+       */
+      children?: ({ isDelayed }: { isDelayed: boolean }) => ReactElement
```
SEOKKAMONI, line 13: I can't say this is the best approach, but it seems it can also be expressed through [ChildrenRenderProps](https://github.com/toss/suspensive/blob/main/packages/jotai/src/utility-types/ChildrenRenderProps.ts).

packages/react/src/Delay.tsx
```diff
@@ -15,11 +25,20 @@ export const Delay = (props: DelayProps) => {
   const defaultProps = useContext(DelayDefaultPropsContext)
   const ms = props.ms ?? defaultProps.ms ?? 0
 
-  const [isDelaying, setIsDelaying] = useState(ms > 0)
-  useTimeout(() => setIsDelaying(false), ms)
+  const [isDelayed, setIsDelayed] = useState(ms <= 0)
+  useTimeout(() => setIsDelayed(true), ms)
```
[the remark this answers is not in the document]
manudeli: Then, Could you give me suggestion code?

packages/react/src/Delay.tsx
manudeli: I'll make return as ReactNode! thanks!

packages/react/src/Delay.tsx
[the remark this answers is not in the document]
SEOKKAMONI: If you believe that setTimeout is not necessary when a negative or zero value is passed to ms in useTimeout, then the following code could be appropriate:

```tsx
import { useEffect } from 'react';
import { usePreservedCallback } from './usePreservedCallback';

export const useTimeout = (fn: () => void, ms: number) => {
  const preservedCallback = usePreservedCallback(fn);

  useEffect(() => {
    if (ms <= 0) {
      preservedCallback();
      return;
    }

    const id = setTimeout(preservedCallback, ms);
    return () => clearTimeout(id);
  }, [preservedCallback, ms]);
};
```
However, if you cannot change the internal code of useTimeout and need to control it from outside, I cannot think of a suitable solution. In such a situation, it seems better to maintain the existing structure.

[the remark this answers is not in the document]
manudeli: This doesn't seem to be necessary. It's not that the callback setTimeout-ed is executed, but rather that when the reference of the state changes, it will cost a lot of computing resources to reflect the state in the dom to the ui.

Even if useTimeout is executed internally in Delay, it doesn't render to reflect it in the dom when the state changes from isDelayted:true -> isDelayted:true, so I don't think it's a problem worth worrying about.

Also, I think the issue of whether to execute the callback received when the ms of useTimeout is 0 will depend on the implementer. So I don't think that if the ms of useTimeout is 0, the callback should not necessarily be executed.

If this is important issue for you, Make next Pull Request to resolve this issue please

SEOKKAMONI (approved):
Good!! 😊

kangju2000 (approved):
😀👍👍👍

PR #1319, opened 2024-10-16
refactor(react): update `usePreservedCallback` signature

dayongkr:
# Overview

Since ₩usePreservedCallback₩ needs to accept functions of all forms, due to contravariance, it’s appropriate to set the function parameter to the lowest possible type(`never`).

Screenshot 2024-10-16 at 11 50 57 AM

However, previously it was set to `unknown`, so when you pass in a function with specified parameter types, TypeScript throws an error.

Screenshot 2024-10-16 at 11 51 16 AM

Initially, I changed it to `never`, but since `never` has a problem in situations like the one above with `implicit any`. Because `implicit any` was inferred as `never`.

Therefore, in the end, I set it to the `any` type.

## PR Checklist

### DIFF
--- a/packages/react/src/hooks/usePreservedCallback.test-d.ts
+++ b/packages/react/src/hooks/usePreservedCallback.test-d.ts
@@ -16,13 +16,7 @@ describe('usePreservedCallback', () => {
     ).toEqualTypeOf<(a: number, b: string) => void>()
 
     expectTypeOf(
-      usePreservedCallback((a: number, b: string) => {
-        if (a > 0) {
-          return b
-        }
-
-        return a
-      })
+      usePreservedCallback((a: number, b: string) => Math.random() > 0.5 ? a : b)
     ).toEqualTypeOf<(a: number, b: string) => number | string>()
     expectTypeOf(
       usePreservedCallback((a) => {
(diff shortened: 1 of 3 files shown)


packages/react/src/hooks/usePreservedCallback.ts
```diff
@@ -1,14 +1,14 @@
...
 
   useIsomorphicLayoutEffect(() => {
     callbackRef.current = callback
   }, [callback])
 
-  return useCallback((...args: unknown[]) => {
+  return useCallback((...args: Parameters<T>) => {
     return callbackRef.current(...args)
   }, []) as T
 }
```
manudeli, line 14: Could you add type test with usePreservedCallback.test-d.ts?

dayongkr: Sure! I added it.

manudeli (approved):
Thanks! 👍

packages/react/src/hooks/usePreservedCallback.test-d.ts
```diff
@@ -0,0 +1,33 @@
...
+    ).toEqualTypeOf<(a: number, b: string) => void>()
+
+    expectTypeOf(
+      usePreservedCallback((a: number, b: string) => {
+        if (a > 0) {
+          return b
+        }
+
+        return a
+      })
```
manudeli, line 25:
```suggestion
      usePreservedCallback((a: number, b: string) => Math.random() > 0.5 ? a : b)
```

dayongkr: Looks more simple!

manudeli:
Resolve eslint error please 🙇

PR #1240, opened 2024-08-31
ci(*): typescript-eslint 8

vr-varad:
# Overview

## PR Checklist

- [ ] I did below actions if need

1. I read the [Contributing Guide](https://github.com/toss/suspensive/blob/main/CONTRIBUTING.md)
2. I added documents and tests.

PR #1326, opened 2024-10-18
feat(eslint-config): typescript-eslint 8

### LINKED ISSUE #1238: [Feature]: typescript-eslint 8
### Package Scope

etc

### Description

@vr-varad Could you do this please?

### Possible Solution

_No response_

### etc.

_No response_

SEOKKAMONI:
# Overview
related #1240

## PR Checklist


manudeli:
cc. @codebutler @saul-atomrigs

configs/tsconfig/base.json
```diff
@@ -15,7 +15,8 @@
     "preserveWatchOutput": true,
     "skipLibCheck": true,
     "strict": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "allowJs": true
```
SEOKKAMONI, line 19:
```cmd
0:0  error  Parsing error: "parserOptions.project" has been provided for @typescript-eslint/parser. The file was not found in any of the provided project(s): eslint.config.mjs
```

By default, tsconfig does not include .js files. This behavior caused the error:
"The file was not found in any of the provided project(s): eslint.config.mjs."
We resolved this issue by adding allowJs: true, which made it possible to treat .js files as part of the project.

ref: https://typescript-eslint.io/troubleshooting/typed-linting#i-get-errors-telling-me--was-not-found-by-the-project-service-consider-either-including-it-in-the-tsconfigjson-or-including-it-in-allowdefaultproject

packages/react-dom/src/test-utils/index.ts
```diff
@@ -8,7 +7,7 @@ declare global {
   var jest: { fn: typeof vi.fn } | undefined
 }
 
-const act = typeof React.act === 'function' ? React.act : DeprecatedReactTestUtils.act
+const act = React.act
```
SEOKKAMONI, line 10: I found that act inside react-dom/test-utils has been deprecated. As mentioned in the documentation, I replaced it with the act from react.

https://react.dev/warnings/react-dom-test-utils

packages/react-query/src/scripts/utils/switchVersion.spec.ts
```diff
@@ -11,8 +12,8 @@ describe('switchVersion', () => {
   let consoleWarnSpy: MockInstance
 
   beforeEach(() => {
-    consoleLogSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
-    consoleWarnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {})
+    consoleLogSpy = vi.spyOn(console, 'log').mockImplementation(noop)
+    consoleWarnSpy = vi.spyOn(console, 'warn').mockImplementation(noop)
```
SEOKKAMONI, line 16: I replaced it with noop.

configs/eslint-config/eslint.config.mjs
```diff
@@ -47,7 +47,7 @@ export default [
       '@typescript-eslint/no-unsafe-member-access': 'warn',
       '@typescript-eslint/no-floating-promises': 'warn',
       '@typescript-eslint/no-unsafe-call': 'warn',
-      '@typescript-eslint/no-throw-literal': 'warn',
+      '@typescript-eslint/only-throw-error': 'warn',
```
SEOKKAMONI, line 50: The @typescript-eslint plugin does not have a no-throw-literal rule. Instead, I replaced it with @typescript-eslint/only-throw-error, which enforces the same rule.

configs/tsconfig/base.json
kangju2000: There are `.js` file extensions in [tseslint.configs.strictTypeChecked](https://github.com/toss/suspensive/pull/1326/files#diff-b30f7315dea69043e473e61637bf36479d298da8a0f1cdac60915633a448fba7R29). Could this be the reason for the parsing error?

kangju2000: I'm wondering if the purpose of ESLint is to lint only TypeScript files. If that's the case, couldn't we remove the JavaScript-related extensions from the [scripts](https://github.com/toss/suspensive/blob/main/configs/eslint-config/package.json#L27)?

```diff
-"ci:eslint": "eslint \"**/*.{js,jsx,cjs,mjs,ts,tsx,cts,mts}\"",
+"ci:eslint": "eslint \"**/*.{ts,tsx,cts,mts}\"",
```

SEOKKAMONI: I respect your opinion, and I believe that excluding checks for .js files in ESLint settings might be a somewhat regrettable choice. ESLint provides various rules that help maintain code conventions, and these rules still hold significant meaning for files that do not use TypeScript. For example, rules like import/order and prettier can greatly enhance code readability and maintain consistency within the project. From this perspective, I think it is more desirable to apply ESLint checks to all files.

SEOKKAMONI: > There are .js file extensions in [tseslint.configs.strictTypeChecked](https://github.com/toss/suspensive/pull/1326/files#diff-b30f7315dea69043e473e61637bf36479d298da8a0f1cdac60915633a448fba7R29). Could this be the reason for the parsing error?

If we remove the .js extension from tseslint.configs.strictTypeChecked, the error will disappear, but this means that ESLint will not be applied to JavaScript files at all, which I think is not an ideal solution.

packages/react/src/Suspensive.spec.tsx
```diff
@@ -13,6 +13,7 @@ const FALLBACK_GLOBAL = 'FALLBACK_GLOBAL'
 describe('<SuspensiveProvider/>', () => {
   it('should provide default ms prop of Delay', async () => {
     render(
+      // eslint-disable-next-line @typescript-eslint/no-deprecated
```
saul-atomrigs, line 16: suggestion) 
This PR adds this `eslint-disable-next-line` in several locations.
How about having it at 'warn' level so we can pass CI and no need to have so many eslint-disables..?

SEOKKAMONI: Using eslint-disable-next-line excessively is indeed an issue, but I’m not entirely sure if changing the rule to 'warn' is the right approach. This change would affect other files, which makes me hesitant.

For example, the most commonly disabled rule is @typescript-eslint/no-deprecated. The places where this rule is disabled involve components like Suspensive and DevMode, which are deprecated and not exposed to users. Since these are very specific cases, I’m concerned that changing the rule to 'warn' might unintentionally affect other parts of the codebase where the rule should still be enforced.

saul-atomrigs: I think the "best" solution is to actually replace the deprecated syntax with new syntax:

Captura de pantalla 2024-10-20 a las 1 14 15 p  m

SEOKKAMONI: This file is intended to test components like Suspensive and SuspensiveProvider. Simply replacing them with DefaultPropsProvider, which doesn’t align with the original purpose, doesn’t seem like a good solution.

I believe that these deprecated files will likely be removed when Suspensive is upgraded to V3. Until then, if there isn’t a better alternative, how about temporarily suppressing the errors with eslint-disable-next-line?

configs/tsconfig/base.json
manudeli: > I believe that excluding checks for .js files in ESLint settings might be a somewhat regrettable choice. ESLint provides various rules that help maintain code conventions, and these rules still hold significant meaning for files that do not use TypeScript.

I agree absolutely!

packages/react/src/Suspensive.spec.tsx
manudeli: This thread is getting long, so to speed up the merge of this PR, I would appreciate it if you could first proceed in the way @SEOKKAMONI did, and then raise this issue later. I will create this as a separate issue.

SEOKKAMONI: Let's continue the discussion further in issue #1331! I’d like to express my gratitude to @saul-atomrigs for leaving a thoughtful comment that helps us dive deeper into this topic!

packages/jotai/src/Atom.test-d.tsx
```diff
@@ -15,72 +15,78 @@ const loadableAtom = loadable(asyncAtom)
...
         {(result) => {
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
           const returnOfJotai = useAtom(countAtom)
           expectTypeOf(result).toEqualTypeOf<typeof returnOfJotai>()
           return <></>
         }}
       </Atom>
-    )
-    ;() => (
+    ))()
```
manudeli, line 27: Cool

packages/react-query-5/src/PrefetchQuery.test-d.tsx
```diff
@@ -1,60 +1,60 @@
+import { queryOptions } from '@tanstack/react-query'
```
manudeli, line 1: 👍

manudeli (approved):
❤️❤️❤️ Thanks!! ❤️❤️❤️

configs/tsconfig/base.json
manudeli: I want to follow this problem as @kangju2000 way. I agree both opinion by you guys, but I see this error! so I made this commit [8408867](https://github.com/toss/suspensive/pull/1326/commits/84088673b707a0a769e06f38042e1ef8bfdc0f1a)
Could you make new Pull Request if you want to check js,jsx,mjs,cjs at that time?

image

packages/react/src/Suspensive.spec.tsx
saul-atomrigs: @SEOKKAMONI thank you and great effort in this PR..!

PR #1327, opened 2024-10-19
refactor(react): improve `<Delay/>`'s performance by simplify useTimeout's code

dayongkr:
# Overview

I understand that using functions like `useEffect` to update state and values received via props is not recommended because it adds unnecessary processes. Especially since there’s only one prop, I don’t think it’s necessary to use it. Therefore, I have removed that part.

If there’s anything I might be missing, please let me know! Thank you.

> And I thought that the `dependencies array` in `useEffect` would be meaningless because `usePreservedCallback(() => {})` is typically used this way, meaning the callback will always receive a new reference value.

> Reference: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

## PR Checklist


manudeli:
https://github.com/juliencrn/usehooks-ts/blob/master/packages/usehooks-ts/src/useTimeout/useTimeout.ts

Thank you for your question about where the performance can be improved. And I need your help.
We just imitate usehooks-ts's useTimeout. We just use that because usehooks-ts is famous and reliable library. so our code is same with its. Could you investigate why they use useIsomorphicLayoutEffect?

gwansikk:
In my opinion, stability is the key consideration. Stability is crucial for libraries, as they need to guarantee consistent results in all scenarios. By using useIsomorphicLayoutEffect in React’s lifecycle process, the intention was likely to ensure the same outcome in every situation.

Without it, unintended behaviors might occur in certain cases (such as abnormal rendering due to specific conditions, asynchronous operations, or interference from parent components). It seems like useIsomorphicLayoutEffect was employed as a stability mechanism to produce consistent results.

manudeli:
> In my opinion, stability is the key consideration. Stability is crucial for libraries, as they need to guarantee consistent results in all scenarios. By using useIsomorphicLayoutEffect in React’s lifecycle process, the intention was likely to ensure the same outcome in every situation.
> 
> Without it, unintended behaviors might occur in certain cases (such as abnormal rendering due to specific conditions, asynchronous operations, or interference from parent components). It seems like useIsomorphicLayoutEffect was employed as a stability mechanism to produce consistent results.

I have same opinion

manudeli:
@gwansikk But we use our useTimeout only in [Delay](https://github.com/toss/suspensive/blob/main/packages/react/src/Delay.tsx) 

Must we use useIsomorphicLayoutEffect? and useEffect? Actually I can't know why we need to useTimeout containing useIsomorpicLayoutEffect. isn't it enough nothing like @dayongkr's implementation?

manudeli:
I think useIsomorphicLayout was used to prevent the ref from changing during the paint process when the callback passed to useTimeout changes. However, Suspensive does not provide useTimeout as a public API, so it can guarantee that the callback passed to useTimeout cannot change.

Then, it would be better to optimize it as implemented by @dayongkr, but I wonder what @gwansikk, @dayongkr thinks. (I think it might take some time. Please reply slowly.)

gwansikk:
> isn't it enough nothing like @dayongkr's implementation?

Exactly. Currently, we are using `usePreservedCallback` internally, and we are not exposing `useTimeout` to users. In that case, I believe following @dayongkr’s approach should be sufficient.

However, could there be a scenario in the future where we use `usePreservedCallback` directly to create APIs? (While I don’t expect this to happen.) Or, could there be a case where a callback passed to `usePreservedCallback` is injected as a component prop in a future interface?

gwansikk:
> However, Suspensive does not provide useTimeout as a public API, so it can guarantee that the callback passed to useTimeout cannot change.

If there’s no chance that it will be exposed externally or directly used in a component and provided as an interface in the future, then I agree with @dayongkr!

manudeli:
I remove unecessary useIsomorphicLayoutEffect too in [3403e8c](https://github.com/toss/suspensive/pull/1327/commits/3403e8cf54c8c73c29f6b657bef818963920105b)

dayongkr:
I tried to find the commits to understand why they were used, but I couldn’t find any reason. As far as I know, `useLayoutEffect` is used to delay the invocation of an effect until the `commit` phase, particularly when using DOM properties. In most cases, it’s hard to observe noticeable benefits from this, and even with a very slow computer(20x slowdown CPU), flickering issues are barely visible.

Anyway, if we are not manipulating the DOM within the effect, I believe there’s no need to delay the invocation until after the commit phase!

manudeli:
I also removed usePreservedCallback too in useTimeout with @dayongkr's implementation in [0f7bdc7](https://github.com/toss/suspensive/pull/1327/commits/0f7bdc733644c8501fe40a9757c8aeefbfde118d)

We don't have to maintain it as internal api when we need to do it. Let's make that!

.changeset/honest-cameras-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@suspensive/react": patch
+---
+
+refactor(react): remove `useIsomorphicLayoutEffect` in `usePreservedCallback`
```
manudeli, line 5:
```suggestion
refactor(react): improve Delay's performance by simplify useTimeout's code
```

manudeli (approved):
@dayongkr This remove many things and improve performance of Delay too. Thanks! 🙇
If @gwansikk approve this Pull Request, I'll merge this. 

@gwansikk but Please feel free to request changes if needed

gwansikk (approved):
@dayongkr Thanks! 👍👍

Let's go!!

dayongkr:
> I tried to find the commits to understand why they were used, but I couldn’t find any reason. As far as I know, `useLayoutEffect` is used to delay the invocation of an effect until the `commit` phase, particularly when using DOM properties. In most cases, it’s hard to observe noticeable benefits from this, and even with a very slow computer(20x slowdown CPU), flickering issues are barely visible.

Some of the information was incorrect, so I have revised it for future reference: Both `useEffect` and `useLayoutEffect` are executed after a commit. However, `useLayoutEffect` runs synchronously before the browser paints, causing a blocking effect that can result in slower performance.

PR #1336, opened 2024-10-26
fix(react): change `defineSuspense` function signature

dayongkr:
# Overview

I made modifications because, when `componentPropsClientOnly` is set to `false`, it returns `Suspense`, but the overloaded function signature indicated that it returns `SuspenseClientOnly`.

## PR Checklist


manudeli (approved):
@dayongkr Thanks! this is concise!

PR #1352, opened 2024-11-13
fix: React Native Playground Configuration

XionWCFM:
# Overview

Configure an environment to run React Native Playground.

## Guide

1. install dependencies 

```
pnpm i
```

2. run project

```
cd examples/react-native-playground
pnpm run start
i
```

3. The following screen will be displayed:

Screenshot 2024-11-13 at 18 08 25

Screenshot 2024-11-13 at 18 12 07

## Test

 built the entire project and confirmed that it was built without a problem.

## Trouble Shooting

React Native is sensitive to cache and nodemodule state.

If it doesn't work well, delete the entire nodemodules and cache folders in the project and try again.

## ETC

I realized that public-hoist-pattern[]=* is not the same as node-linker=hoisted, but rather the same as setting shamefully-hoist to true.

I set node-linker=hoisted in another project and experienced next15's build failing during the CI process.

By default, pnpm manages dependencies strictly so that only specified dependencies are used.

However, if you set this setting, you will give up strict dependency management and manage dependencies similar to the method of older generation package managers such as yarn and npm.

Should we give up the functionality of pnpm? I wish there was another way to solve this problem.

I think I can use it until I find another way. shamefully-hoist and public-hoist-pattern[]=*


.npmrc
```diff
@@ -1 +1,2 @@
 prefer-workspace-packages=true
+public-hoist-pattern[]=*
```
manudeli, line 2:
```suggestion
public-hoist-pattern[]=*

```

examples/react-native-playground/.gitignore
```diff
@@ -0,0 +1,6 @@
+
+# @generated expo-cli sync-2b81b286409207a5da26e14c78851eb30d8ccbdb
+# The following patterns were generated by expo-cli
+
+expo-env.d.ts
+# @end expo-cli
```
manudeli, line 6:
```suggestion
# @end expo-cli

```

examples/react-native-playground/package.json
```diff
@@ -1,24 +1,53 @@
...
-    "@expo/webpack-config": "catalog:",
-    "@types/react": "catalog:react18"
-  }
+    "@babel/core": "^7.20.0",
+    "@types/react": "catalog:react18",
+    "jest": "catalog:",
+    "jest-expo": "catalog:",
+    "typescript": "~5.3.3"
+  },
+  "packageManager": "pnpm@9.12.0+sha512.4abf725084d7bcbafbd728bfc7bee61f2f791f977fd87542b3579dcb23504d170d46337945e4c66485cd12d588a0c0e570ed9c477e7ccdd8507cf05f3f92eaca"
```
manudeli, line 52:
```suggestion
```

XionWCFM: That was my personal setting and I forgot to delete it.

I'll add a commit to delete that setting. Thank you!

manudeli: Thanks!

XionWCFM: I edited this part!

manudeli:
@XionWCFM Could you resolve https://github.com/toss/suspensive/actions/runs/11814909318/job/32915046140?pr=1352 this ci error please?

XionWCFM:
In my other projects, when I set hoisted, a problem occurred in another next15 (probably not solving the react version properly).

Please understand that we may create a few more commits for testing in CI.

XionWCFM:
Could you please test this commit?  @manudeli 

If this commit doesn't work, you may have to use the hacky solution I originally suggested for CI.

(use a different .npmrc locally only)

.npmrc
manudeli, line 2:
```suggestion
public-hoist-pattern[]=*

```

XionWCFM:
The build appears to have been successful.

I thought that globing node-linker = hoisted and public-hoist-pattern=* seemed to be semantically equivalent.

However, node-linker=hoisted affects the next 15 projects in your project and the public-hoist-pattern task.

Can you guess the cause of this? haha..

manudeli (approved):
Thank you for your research and PR❤️
Thanks to your help, we will be able to support @suspensive/react-native more quickly.

PR #1356, opened 2024-11-14
feat(react-native-playground): remove shamefully-hoist and expo 52 version upgrade

XionWCFM:
# Overview

With the addition of support for pnpm monorepo in expo 52, you can now disable "shamefully hoist" and "node-linker=hoisted".

[expo sdk 52 release](https://expo.dev/changelog/2024/11-12-sdk-52)

Accordingly, we upgrade the existing react-native-playground's expo version to 52 and remove "public-hoist-pattern[]=*".

As a result of testing, if you use the monorepo pattern without hoisting, it works fine locally, but causes problems in EAS remote builds.

However, Playground is not expected to require EAS remote build. 

I guess the benefits of not using the hoisting pattern are greater.

## Test

I confirmed that the IOS Simulator operates locally and that the entire build process operates.

## Trouble Shooting

During the migration process, I often failed to start the expo app due to the cache of metro or packagemanager.

Most problems are caused by mismatched ReactNative versions.

If you are having this problem try

```
pnpm store prune
pnpm i
```

[pnpm store prune docs](https://pnpm.io/next/cli/store#prune)

## PR Checklist


.npmrc
```diff
@@ -1,2 +1 @@
-prefer-workspace-packages=true
-public-hoist-pattern[]=*
+prefer-workspace-packages=true
```
manudeli, line 1:
```suggestion
prefer-workspace-packages=true

```

manudeli (approved):
So cool❤️ Thanks! Best environment to develop @suspensive/react-native

PR #1363, opened 2024-11-19
chore(.github/actions/pnpm-setup-node): remove version field in with of pnpm/action-setup

sukvvon:
# Overview

removed the version specified in the with field of pnpm/action-setup. [when the pnpm version is already defined in the packageManager field of package.json, pnpm/action-setup can automatically recognize and use it](https://github.com/pnpm/action-setup/tree/v4/?tab=readme-ov-file#version). duplicating the version setting adds unnecessary complexity and increases maintenance overhead without added benefit.

by relying on the version specified in packageManager, this change simplifies the configuration and ensures consistency.

## PR Checklist

### DIFF
--- a/.github/actions/pnpm-setup-node/action.yml
+++ b/.github/actions/pnpm-setup-node/action.yml
@@ -4,10 +4,8 @@ runs:
   using: composite
   steps:
     - uses: pnpm/action-setup@v4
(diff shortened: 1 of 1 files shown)


manudeli (approved):
👍

PR #1365, opened 2024-11-19
feat(react, react-query-4): add missing JSDoc for react and react-query-4

kangju2000:
# Overview

I add JSDoc comments to `@suspensive/react` and `@suspensive/react-query-4`.

## PR Checklist

### DIFF
--- a/packages/react/src/ClientOnly.tsx
+++ b/packages/react/src/ClientOnly.tsx
@@ -6,4 +6,8 @@ export interface ClientOnlyProps {
   fallback?: ReactNode
 }
 
+/**
+ * This component ensures its children are only rendered on the client-side.
+ * @see {@link https://suspensive.org/docs/react/ClientOnly Suspensive Docs}
+ */
 export const ClientOnly = ({ children, fallback }: ClientOnlyProps) => <>{useIsClient() ? children : fallback}</>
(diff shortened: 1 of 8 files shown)


packages/react/src/ClientOnly.tsx
```diff
@@ -6,4 +6,8 @@ export interface ClientOnlyProps {
   fallback?: ReactNode
 }
 
+/**
+ * This component ensures its children are only rendered on the client-side.
+ * @see {@link https://suspensive.org/docs/react/ClientOnly Suspensive Docs}
```
kangju2000, line 11: Which one do you prefer between `https://suspensive.org/docs` and `https://suspensive.org/en/docs`

manudeli: https://suspensive.org/docs because https://suspensive.org/docs will automatically redirect user's page to proper language page if user use any language

manudeli (approved):
❤️

packages/react/src/ClientOnly.tsx
kangju2000: It would be good to standardize it to links without 'en' since they're not unified!

manudeli: Oh,, I merge this Pull Request. could you do that in next pull request please?

kangju2000: Yes, I'll raise a PR to standardize the jsDoc overall!

PR #1338, opened 2024-10-29
feat(codemods): add @suspensive/codemods package

gwansikk:
# Overview

I have designed a codemods CLI that allows you to select and execute various transforms (codemods).

Currently, only the `tanstack-query-import` codemod has been implemented, but we plan to add more codemods that facilitate version upgrades and make codebase maintenance easier. For example:

- suspensive v1 → v2 break change codemod
- suspensive v2 → v3 break change  codemod

## tanstack-query-import

> Migrate imports to @tanstack/react-query in @suspensive/react-query (using @tanstack/react-query@5)

I’ve implemented the `tanstack-query-import` codemod. This tool migrates deprecated APIs from `@suspensive/react-query-5` to `@tanstack/react-query` by simply replacing imports. It helps facilitate an easy migration when updating from `@suspensive/react-query-4` to `@suspensive/react-query-5`.

### Preview

https://github.com/user-attachments/assets/4140416d-fcf7-437f-bb24-35f918afcd70

## PR Checklist

- [x] I did below actions if need

1. I read the [Contributing Guide](https://github.com/toss/suspensive/blob/main/CONTRIBUTING.md)
2. I added documents and tests.

PR #1370, opened 2024-11-22
refactor(tsup): update script options

gwansikk:
# Overview

details: https://github.com/toss/suspensive/pull/1338#discussion_r1822459253

- Standardized tsup script options.
- Narrowed down the utils folder in react-query and codemods. (The react-query project structure has been updated accordingly.)
- I verified that it works correctly locally using pnpm pack.

## PR Checklist


packages/react-query/src/bin/postinstall.ts
```diff
@@ -0,0 +1,22 @@
...
+  } else if (version.startsWith('5.')) {
+    switchVersion(5)
+  } else {
+    console.warn('[@suspensive/react-query]', `version v${version} is not supported.`)
+  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+} catch (error) {
+  console.error('[@suspensive/react-query]', 'not fond @tanstack/react-query')
+  process.exit(1)
+}
```
manudeli, line 22: Original code
```ts
import { getTanStackReactQueryPackageJson } from './utils/package'
import { switchVersion } from './utils/switchVersion'

const { version } = getTanStackReactQueryPackageJson()

if (version.startsWith('4.')) {
  switchVersion(4)
} else if (version.startsWith('5.')) {
  switchVersion(5)
} else {
  console.warn('[@suspensive/react-query]', `version v${version} is not supported.`)
}
```

I suggest below code.

```suggestion
const { version } = getTanStackReactQueryPackageJson()

if (!version) {
  throw new Error('[@suspensive/react-query]', 'not found version of @tanstack/react-query')
}

if (version.startsWith('4.')) {
  switchVersion(4)
} else if (version.startsWith('5.')) {
  switchVersion(5)
} else {
  console.warn('[@suspensive/react-query]', `version v${version} is not supported.`)
}
```

if you want use try catch, you could make specific error constructor to specify that caught error is for not found version

```suggestion
class NotFoundVersionError extends Error {
  message: string = 'Not found version'
}

try {
  const { version } = getTanStackReactQueryPackageJson()

  if (!version) {
    throw new NotFoundVersionError()
  }

  if (version.startsWith('4.')) {
    switchVersion(4)
  } else if (version.startsWith('5.')) {
    switchVersion(5)
  } else {
    console.warn('[@suspensive/react-query]', `version v${version} is not supported.`)
  }
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
} catch (error) {
  // but in my opinion, this is not good idea
  if(error instanceof NotFoundVersionError){ // specify error
    console.error('[@suspensive/react-query]', error.message)
  }
  process.exit(1)
}
```

manudeli (changes requested):

packages/react-query/src/bin/postinstall.ts
gwansikk: My intention was to simplify the postinstall script. Ultimately, I designed it so that if an error occurs at runtime (e.g., `node ./dist/bin/postinstall.cjs || echo 'please reinstall @suspensive/react-query'`), the error would propagate as intended. The goal was to consistently inform the user of an appropriate solution through a clear message.

Therefore, I believe this file does not cause any inconvenience to the user due to node(runtime) errors, and I will revert it to its previous state (this was my mistake).

manudeli (approved):
Simple than before 👍

gwansikk:
> @gwansikk Test after new version release please 🙏 I'll release this now

I'm ready!

PR #1383, opened 2024-12-03
refactor(react-query): improve cli utils

### LINKED ISSUE #1292: [Feature]: test coverage 100% for @suspensive/react-query
### Package Scope

@suspensive/react-query

### Description

image

### Possible Solution

_No response_

### etc.

_No response_

gwansikk:
# Overview

closed: https://github.com/toss/suspensive/issues/1292

- modified the cli utils function to throw errors.
- add tests for react-query (100%)

image

## PR Checklist


packages/react-query/src/bin/utils/copy.ts
```diff
@@ -10,7 +10,7 @@ export function copy(version: number) {
     return false
   }
 
-  files.forEach((file) => {
+  for (const file of files) {
```
gwansikk, line 13: ref: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md

This may not have a significant impact on our project, but I agree with the intent and would like to follow it.

packages/react-query/src/bin/utils/package.ts
```diff
@@ -1,15 +1,13 @@
...
+export function loadModule<T>(name: string): T {
   try {
     // eslint-disable-next-line @typescript-eslint/no-require-imports
-    return { exports: require(name) as T, isSuccess: true }
+    return require(name) as T
   } catch {
-    return { exports: undefined, isSuccess: false }
+    throw new Error(`${name} is not found.`)
   }
 }
```
gwansikk, line 12: Previously, loadModule did not throw errors directly but instead followed a structure where the CLI (node) would terminate at the point where the function was used. However, this approach made testing difficult, as mocking require in Node.js is not possible. For this reason, we decided to modify the utils function to throw errors directly, as it allows for more reliable testing.

P.S. Once this PR is merged, I plan to implement a global error handler for the CLI and custom error classes in the next version.

ref: https://github.com/tj/commander.js?tab=readme-ov-file#override-exit-and-output-handling

manudeli (approved):
Sorry for late review! Thanks👍

PR #1409, opened 2025-01-06
feat(react): prevent recursive exposing fallback when fallback throw error

manudeli:
🚧 This PR could make a huge difference, so please help us make sure this change gets a very thorough review. 🚧 

```tsx
const Throw = {
  Error: ({ message, after = 0, children }: PropsWithChildren<{ message: string; after?: number }>) => {
    const [isNeedThrow, setIsNeedThrow] = useState(after === 0)
    if (isNeedThrow) {
      throw new Error(message)
    }
    useTimeout(() => setIsNeedThrow(true), after)
    return <>{children}</>
  }
}

const Example = () => (
  <ErrorBoundary fallback={() => <>This is expected</>}>
    <ErrorBoundary
      fallback={() => (
        <Throw.Error message={ERROR_MESSAGE} after={100}>
          ErrorBoundary's fallback before error
        </Throw.Error>
      )}
    >
      <Throw.Error message={ERROR_MESSAGE} after={100}>
        ErrorBoundary's children before error
      </Throw.Error>
    </ErrorBoundary>
  </ErrorBoundary>
)
```

## Problem: ErrorBoundary's fallback can't be treated by parent ErrorBoundary

 
Thrown Error in fallback will be caught by ErrorBoundary self and then expose fallback recursively 🥲
1. ErrorBoundary's children before error
2. ErrorBoundary's fallback before error
3. ErrorBoundary's fallback before error
4. ErrorBoundary's fallback before error
5. ... expose fallback self recursively ... 

## Solution: When we meet thrown error in fallback of ErrorBoundary wrap it as InternalFallbackError, re-throw InternalFallbackError.fallbackError, if it is InternalFallbackError

Thrown Error in fallback will be caught by parent ErrorBoundary 👍
1. ErrorBoundary's children before error
2. ErrorBoundary's fallback before error
3. This is expected 

## PR Checklist

5. I added documents and tests.


gwansikk (approved):
I agree with this change. It aligns with the same mechanism used by React Suspense, [react-error-boundary](https://github.com/bvaughn/react-error-boundary) (both propagate issues to the parent when a problem occurs in the fallback).

However, I’m concerned that this change could significantly impact existing users. I believe it qualifies as a breaking change, as updating might result in different behavior for previous users.

Additionally, looking at it differently, I think a looping fallback could be an optional feature. Why must errors within a fallback propagate to the parent fallback? What if this behavior were made configurable? (this is just my opinion.)

fe-dudu:
[react-query docs | useSuspenseQuery](https://tanstack.com/query/v5/docs/framework/react/guides/suspense#throwonerror-default).

@manudeli I wish `<Throw.Error/>` could be changed.

because I want to see the `ErrorBoundary` if an error occurs when executing `invalidateQueries` after fetching data to `useSuspenseQuery` or `useSuspenseQueries`.

After reading the react-query docs, I made a component like this. (before I saw this PR)
```js
interface Props {
  isFetching: boolean;
  error: Error | null;
  children: React.ReactNode;
}

export default function ThrowError({ isFetching, error, children }: Props) {
  if (error && !isFetching) {
    throw error;
  }

  return <>{children}</>;
}
```

```js
<SuspenseQuery {...itemsQueryOptions()}>
  {({ data: items , isFetching, error }) => (
    <ThrowError isFetching={isFetching} error={error}>
      <QueryClientConsumer>
        {queryClient => (
          <>
            {items.map(item => (
              <ItemCard
                key={item.id}
                item={item}
                onClick={() => {
                  queryClient.invalidateQueries(itemsQueryOptions());
                }}
              />
            ))}
          </>
        )}
      </QueryClientConsumer>
    </ThrowError>
  )}
</SuspenseQuery>
```
I think `ThrowError` will be necessary to throw the error to the parent when the error occurs in the refetch caused by `invalidateQueries`.

I hope the `<Throw.Error/>` you made functions similarly to my `<ThrowError/>`.

---

The link below is a website that has an API error with a 50% probability. When you click on the card, `invalidateQueries` works.

https://suspensive-ex.vercel.app
https://github.com/fe-dudu/suspensive-ex

manudeli:
I'm considering this as Suspensive v3

manudeli:
> Hi, I'm a big fan of Suspensive. I have a suggestion 🙌
> 
> @manudeli I wish `<Throw.Error/>` could be changed.
> 
> because I want to see the `ErrorBoundary` if an error occurs when executing `invalidateQueries` after fetching data to `useSuspenseQuery` or `useSuspenseQueries`.
> 
> After reading the [react-query docs | useSuspenseQuery](https://tanstack.com/query/v5/docs/framework/react/guides/suspense#throwonerror-default), I made a component like this. (before I saw this PR)
> 
> ```js
> interface Props {
>   isFetching: boolean;
>   error: Error | null;
>   children: React.ReactNode;
> }
> 
> export default function ThrowError({ isFetching, error, children }: Props) {
>   if (error && !isFetching) {
>     throw error;
>   }
> 
>   return <>{children}</>;
> }
> ```
> 
> ```js
> <SuspenseQuery {...itemsQueryOptions()}>
>   {({ data: items , isFetching, error }) => (
>     <ThrowError isFetching={isFetching} error={error}>
>       <QueryClientConsumer>
>         {queryClient => (
>           <>
>             {items.map(item => (
>               <ItemCard
>                 key={item.id}
>                 item={item}
>                 onClick={() => {
>                   queryClient.invalidateQueries(itemsQueryOptions());
>                 }}
>               />
>             ))}
>           </>
>         )}
>       </QueryClientConsumer>
>     </ThrowError>
>   )}
> </SuspenseQuery>
> ```
> 
> I think `ThrowError` will be necessary to throw the error to the parent when the error occurs in the refetch caused by `invalidateQueries`.
> 
> I hope the `<Throw.Error/>` you made functions similarly to my `<ThrowError/>`.
> 
> The link below is a website that has an API error with a 50% probability. When you click on the card, `invalidateQueries` works.
> 
> [suspensive-ex.vercel.app](https://suspensive-ex.vercel.app) [fe-dudu/suspensive-ex](https://github.com/fe-dudu/suspensive-ex?rgh-link-date=2025-01-09T17%3A21%3A58Z)

I agree that it would be good to discuss whether such an interface is needed, but I think adding such an interface would be better discussed somewhere else than this Pull Request. Make issue or pull request please.

manudeli:
> Additionally, looking at it differently, I think a looping fallback could be an optional feature. Why must errors within a fallback propagate to the parent fallback? 

@gwansikk I think developers will expect a similar implementation, as [the official React documentation introduces ErrorBoundary as an interface that replaces try catch](https://react.dev/reference/react/use#dealing-with-rejected-promises).

image

### try catch

```tsx
try {
  try {
    throw new Error('throwed in try');
  } catch {
    throw new Error('throwed in catch inside'); // infinite loop does not occur
  }
} catch (e) {
  console.log(e); // caught
}
```

### Expected ErrorBoundary

```tsx
<ErrorBoundary fallback={({ error }) => console.log(error)}> // I thought the throw that occurred in fallback(catch) should be caught at the top.
  <ErrorBoundary fallback={() => { throw new Error('throwed in fallback(catch)') }}> // I thought an infinite loop shouldn't happen
    {createElement(() => {
      throw new Error('throwed in children(try)')
    })}
  </ErrorBoundary>
</ErrorBoundary>
```

> What if this behavior were made configurable(options)? (this is just my opinion.)

I think configurable looping fallback is not good idea. I think the more settings a library user has to know, the harder it becomes to understand. Also, exposing fallbacks loop doesn't have the advantage of being configurable.

Since this is a breaking change, it seems like it would be easier to migrate by providing a prop that guarantees the same behavior as the existing ErrorBoundary. However, it would be nice if this prop had @deprecated in jsdoc.

gwansikk:
> Since this is a breaking change, it seems like it would be easier to migrate by providing a prop that guarantees the same behavior as the existing ErrorBoundary. However, it would be nice if this prop had `@deprecated` in jsdoc.
> 
> I think it would be better to continue this in another Pull Request/issue to provide a configurable prop to make it behave like the existing ErrorBoundary, so I'll merge it first. 👍

Great, I completely understand and agree!

PR #1410, opened 2025-01-10
fix(react): prevent rendering fallback if not shouldCatch error

manudeli:
issued by @LineGu @lucas0530

# Prevent rendering fallback if not shouldCatch error

```tsx
import { ErrorBoundary } from '@suspensive/react'
import { Throw } from '~/components/Throw'

export default function page() {
  return (
    <ErrorBoundary fallback={() => <>root fallback</>}>
      <ErrorBoundary
        shouldCatch={(error) => error.message !== 'children error message'}
        fallback={() => {
          console.log("child ErrorBoundary's fallback")
          return <>child ErrorBoundary's fallback</>
        }}
      >
        <Throw.Error message="children error message" after={2000}>
          before throw Error
        </Throw.Error>
      </ErrorBoundary>
    </ErrorBoundary>
  )
}
```

fallback should be exposed only when error is catched by ErrorBoundary
But in this case fallback was rendered and `console.log("child ErrorBoundary's fallback")` was excuted. so we fixed it

## PR Checklist


LineGu (approved):
👍

manudeli:
@gwansikk @kangju2000 Check this PR together please. this is important fix in my opinion

gwansikk (approved):
I agree with this change. If `shouldCatch` is false, I believe the fallback itself should not be executed.

In my opinion, the fallback of an ErrorBoundary is meant to act as a substitute for an error. Therefore, if error capturing has not occurred through shouldCatch, the ErrorBoundary should not treat it as an error, and the fallback itself should not be executed.

PR #1414, opened 2025-01-14
fix(react): re-throw SuspensiveError before rerender

manudeli:
# Overview

componentDidCatch will be triggered after render. so I move code re-throwing SuspensiveError from componentDidCatch to render

## PR Checklist


gwansikk (approved):
I agree, LTGM 👍👍

kangju2000 (approved):
👍👍👍

PR #1452, opened 2025-03-02
feat(react): with as method

### LINKED ISSUE #1451: [Feature]: wrap as method
### Package Scope

@suspensive/react

### Description

Add withXXX (withSuspense, withErrorBoundary, withErrorBoundaryGroup)

1. Easier to understand. When explaining v2's wrap to many people, it's not always understood at once. We don't need to know the builder.
2. Since wrap internally depends on Suspense, ErrorBoundary, and ErrorBoundaryGroup, unused functions are included in the bundle, increasing the bundle size.

### Possible Solution

_No response_

### etc.

_No response_

manudeli:
close #1451 

# with as method

There were two reasons why withXXX was inconvenient ( https://github.com/toss/suspensive/pull/270 )
1. It was inconvenient to read in the case of nested hoc by receiving xxxProps as the second argument
2. I thought it was difficult to read in the case of nested Suspense, ErrorBoundary, etc.

## v1 withXXX
```tsx
export default withErrorBoundary(
  withSuspense(
    () => {
     /**
      * 
      * 
      * 
      * Many implementation will be here
      * 
      * 
      *  
      */

      return 'success'
    },
    { fallback: 'loading…' } // far position with name of withSuspense
  ),
  { fallback: () => 'error' } // far position with name of withErrorBoundary
)
```

## v2 wrap.XXX
```tsx
export default wrap
  .ErrorBoundary({ fallback: () => 'error' }) // Good to read
  .Suspense({ fallback: 'loading…' }) // Good to read
  .on(() => {  // But users need to know builder pattern
     /**
      * 
      * 
      * 
      * Many implementation will be here
      * 
      * 
      *  
      */

      return 'success'
    })
```

Because of these two points, I thought that the code would be readable if I made it with wrap builder, but when library users actually used v2 wrap, they didn't understand how to use it nestedly or how to use wrap well.

So I want to change it to a with method for each component like below code so that it receives props first, although it's simple.
1. I want to make props are received first.
2. I don't want it to be a builder.

## v3 XXX.with
```tsx
export default ErrorBoundary.with({ fallback: () => 'error' },  // Good to read
  Susepnse.with({ fallback: 'loading…' }, () => {  // Good to read
   /**
    * 
    * 
    * 
    * Many implementation will be here
    * 
    * 
    *  
    */

    return 'success'
  })
)
```

## PR Checklist


packages/react/src/ErrorBoundaryGroup.tsx
```diff
@@ -33,29 +36,35 @@ export interface ErrorBoundaryGroupProps extends PropsWithChildren {
...
-      ErrorBoundaryGroup.displayName = 'ErrorBoundaryGroup'
-    }
-
-    return ErrorBoundaryGroup
-  })(),
+    return <ErrorBoundaryGroupContext.Provider value={value}>{children}</ErrorBoundaryGroupContext.Provider>
+  },
   {
+    displayName: 'ErrorBoundaryGroup',
+    wrap: <TProps extends ComponentProps<ComponentType> = Record<string, never>>(
```
gwansikk, line 56: How about using `with`? The official documentation for React 16-19 uses the `with` prefix in examples. Because of this, people may find `with` more familiar when recognizing HOCs. (This is just my opinion.)

gwansikk (approved):
I agree. I expect that this interface will resolve the previous issue.

packages/react/src/ErrorBoundaryGroup.tsx
manudeli: I agree with your opinion, We don't have to keep name as wrap.

.changeset/moody-keys-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@suspensive/react": major
+---
+
+feat(react): wrap as method
```
manudeli, line 5:
```suggestion
feat(react): with as method
```

manudeli:
@gwansikk @kangju2000 I will merge first and reflect again if there are any issues.

Issue #1442, opened 2025-02-19
[Feature]: change react query v4 network mode default to always


gwansikk:
### Package Scope

@suspensive/react-query-4

### Description

changed the default network mode to "always" in react query v4.

manudeli:
Let's do this feature as v3 BREAKING CHANGE

manudeli:
@gwansikk Could you make Pull Request for this issue please?
I think this should be tested by vitest in @suspensive/react-query-v4

gwansikk:
> @gwansikk Could you make Pull Request for this issue please?
> I think this should be tested by vitest in @suspensive/react-query-v4

Of course. I’ll give it a try by the end of the day.

manudeli:
@gwansikk Let's do this issue urgently!

gwansikk:
> [@gwansikk](https://github.com/gwansikk) Let's do this issue urgently!

This task will be completed by the end of today! I'm working on it.

manudeli:
Filling migration guide is also important task. Keep it mind too please

PR #1458, opened 2025-03-03
feat(react-query-4): set network mode to always for suspense

gwansikk:
# Overview

closed: #1442 

In React Query v4, the `OnlineManager` checks the client's network status using `navigator.onLine`. However, this can sometimes cause bugs in Chrome (Chromium-based browsers), where the network is incorrectly detected as offline even when it is online.  

To minimize this issue, React Query v5 no longer relies on `navigator.onLine`.  

In React Query v4, to minimize this issue, **networkMode is set to always in Suspense Hooks**. This change ensures that TData in Suspense remains guaranteed.

ref:

- https://github.com/TanStack/query/blob/v4/packages/query-core/src/onlineManager.ts#L92
- https://github.com/TanStack/query/discussions/4753
- https://issues.chromium.org/issues/338514113

## PR Checklist


gwansikk:
It's important to note that in React Query 4, the `networkMode` for Suspense Hooks has been fixed to `always`, but I don't believe this is a high priority at the moment. I will manage the documentation work by creating a separate issue.

manudeli (approved):
Cool! We can get perfact return type of useSuspenseQuery finally

PR #1518, opened 2025-04-12
feat(react-dom): onInView, onInViewEnd

manudeli:
# Overview

## PR Checklist


docs/suspensive.org/src/content/ko/docs/react-query/PrefetchQuery.mdx
```diff
@@ -16,9 +16,9 @@ const PostsPage = () => {
 
   return posts.map((post) => (
     <InView>
-      {({ ref, isInView }) => (
+      {({ ref, inView }) => (
```
kangju2000, line 19: 👍👍

packages/react-dom/src/useInView.tsx
```diff
@@ -44,6 +48,11 @@ export function useInView({
       (inView, entry) => {
         setState({ inView, entry })
         onChangeRef.current?.(inView, entry)
+        if (inView) {
+          onInView?.(entry)
+        } else {
+          onInViewEnd?.(entry)
+        }
```
kangju2000, line 55: I’m wondering if onInViewEnd gets triggered on mount when the element is initially outside the viewport?

manudeli: Oh... You save me! Thanks

manudeli: I fixed it

packages/react-dom/src/InView.tsx
```diff
@@ -2,8 +2,26 @@ import { type InViewOptions, useInView } from './useInView'
...
+        useInView({
+          ...options,
+          onChange: (inView, entry) => {
+            options.onChange?.(inView, entry)
+            if (inView) {
+              options.onInView?.(entry)
+            } else {
+              options.onInViewEnd?.(entry)
+            }
+          },
```
kangju2000, line 22: Wouldn’t onInViewEnd be triggered when the InView component gets rendered but isn’t in the viewport? 👀

manudeli: When I test in visualization, You're right. 
I tried many things but it doesn't work..! If you have any idea what to do, please help me.

kangju2000: @manudeli How about this?

```ts
export function InView({ children, ...options }: InViewProps) {
  const hasBeenInViewRef = useRef(false)

  return (
    <>
      {children(
        useInView({
          ...options,
          onChange: (inView, entry) => {
            options.onChange?.(inView, entry)
            if (inView) {
              hasBeenInViewRef.current = true
              options.onInView?.(entry)
            } else if (hasBeenInViewRef.current) {
              options.onInViewEnd?.(entry)
            }
          },
        })
      )}
    </>
  )
}
```

manudeli: You're king 👑 I use your way in [d9274e4](https://github.com/toss/suspensive/pull/1518/commits/d9274e456a48db711ea0ad6a6d211ca7e20c951c)

manudeli:
@kangju2000 Could you review again. In my opinion, this apis is made by you

kangju2000 (approved):
LGTM 👍👍

PR #1522, opened 2025-04-13
refactor(react-query): update version check logic to use switch statement

m2na7:
# Overview

- Replaced the if-else statement with a switch statement to improve the version check logic.

This approach will be more scalable in the future as additional versions of react-query are introduced !

## PR Checklist


manudeli:
@m2na7 fix eslint error please

m2na7:
@manudeli Fixed eslint error by adding the missing line break in `postinstall.ts`

gwansikk (commented):
Thank you for your contribution. I'm not sure what advantages this change will bring. What are the benefits? (I wish you had emphasized them a bit more.)

m2na7:
Thanks for your comment! @gwansikk

I went with a `switch` statement mainly to make the code easier to maintain as we support more versions down the line. When there are lots of conditions, `if-else` chains can get a bit hard to read, whereas `switch` keeps things cleaner and more scalable.

Also, by extracting the major version once with `version.split('.')[0],` we avoid redundant logic. Plus, using a `switch` helps reduce eye movement while reading the code, making it easier to scan through the logic quickly. (The current version, with its long horizontal structure, can disrupt the flow of the eyes.)

gwansikk (approved):
I agree! Thank you for your contribution.

packages/react-query/src/bin/postinstall.ts
```diff
@@ -4,10 +4,16 @@ import { switchVersion } from './utils/switchVersion'
 
 const { version } = getTanStackReactQueryPackageJson()
 
-if (version.startsWith('4.')) {
-  switchVersion(4)
-} else if (version.startsWith('5.')) {
-  switchVersion(5)
-} else {
-  logger.error(`version v${version} is not supported.`)
+const major = version.split('.')[0]
```
manudeli, line 7: I'm curious if check below case well

```json
"@tanstack/react-query": "5"
```

m2na7: `split('.')[0]` splits the version string by the `.` character and returns the first element. 
However, if the version string doesn't contain a `.`, `split('.')` will return an array containing just the string itself.
For example, if the version is `"5"`, `split('.')` returns `["5"]`, and the first element of that array is `"5"`.

Therefore, `split('.')[0]` works fine even when there is no `.` in the version string, and it simply returns `"5"`

gwansikk:
Can you check the lint errors and fix them?

packages/react-query/src/bin/postinstall.ts
manudeli:
```json
"@tanstack/react-query": "^5"
```

work In this case? cc. @gwansikk

manudeli: In my opinion, We can take npm's implementation( https://github.com/npm/node-semver )

manudeli:
> Can you check the lint errors and fix them?

I don't know why getting this ci error

packages/react-query/src/bin/postinstall.ts
gwansikk: I agree. I was missing that point. It seems necessary to manage it more safely through node-semver.

```typescript
semver.major(version)
```

Additionally, should we also consider npm-dist-tag? (e.g. `latest`)

ret: https://docs.npmjs.com/cli/v11/commands/npm-dist-tag

manudeli: @m2na7 @gwansikk Can you guys review my Pull Request? https://github.com/toss/suspensive/pull/1533
I added you guys as co-author! Thanks!

manudeli: Ah,, we can meet case like ^4.2.2 because version field of @tanstack/react-query's package.json won't write like that

manudeli (approved):

manudeli:
@m2na7 Thanks!

PR #1523, opened 2025-04-16
test(react): use fake timers strictly for Delay.spec.tsx

manudeli:
# Overview

## PR Checklist

### DIFF
--- a/packages/react/src/Delay.spec.tsx
+++ b/packages/react/src/Delay.spec.tsx
@@ -1,4 +1,4 @@
-import { render, screen, waitFor } from '@testing-library/react'
+import { render, screen } from '@testing-library/react'
 import ms from 'ms'
 import { DefaultProps, DefaultPropsProvider } from './DefaultProps'
 import { Delay } from './Delay'
@@ -6,7 +6,7 @@ import { Message_Delay_ms_prop_should_be_greater_than_or_equal_to_0, SuspensiveE
 import { CustomError, TEXT } from './test-utils'
 
 beforeEach(() => {
-  vi.useFakeTimers({ shouldAdvanceTime: true })
+  vi.useFakeTimers()
 })
 
 afterEach(() => {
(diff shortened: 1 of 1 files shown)


packages/react/src/Delay.spec.tsx
```diff
@@ -16,10 +16,11 @@ afterEach(() => {
 
 describe('<Delay/>', () => {
   it('should render the children after the delay', async () => {
-    render(<Delay ms={ms('0.1s')}>{TEXT}</Delay>)
-
+    render(<Delay ms={ms('0.5s')}>{TEXT}</Delay>)
     expect(screen.queryByText(TEXT)).not.toBeInTheDocument()
-    await waitFor(() => expect(screen.queryByText(TEXT)).toBeInTheDocument())
```
manudeli, line 22: strictly time check by fake timers

manudeli:
@ssi02014 Maybe you like this change
You made this test code file for `<Delay/>` at first time

ssi02014 (approved):

packages/react/src/Delay.spec.tsx
```diff
@@ -16,10 +16,11 @@ afterEach(() => {
...
+    await vi.advanceTimersByTimeAsync(ms('0.5s'))
+    await vi.advanceTimersByTimeAsync(0)
```
ssi02014, line 22: Should I call `advanceTimersBy TimeAsync(0)` here to get strict type checking? 😲

manudeli: If I remove `await vi.advanceTimersByTimeAsync(0)`, React renderer won't make new UI to test it
I don't really like this part either, but I think this is the right way to do it. I think it's because React uses timers internally.

ssi02014: @manudeli I understand. Thank you for your hard work.

PR #1528, opened 2025-04-17
feat(react-dom): is prefix for InView

manudeli:
BREAKING CHANGE

# Overview

inView -> isInView

## PR Checklist

### DIFF
--- a/.changeset/six-impalas-remain.md
+++ b/.changeset/six-impalas-remain.md
@@ -1,0 +1,5 @@
+---
+"@suspensive/react-dom": major
+---
+
+feat(react-dom): is prefix, ms suffix for InView
--- a/.changeset/six-impalas-remain.md
+++ b/.changeset/six-impalas-remain.md
@@ -2,4 +2,4 @@
 "@suspensive/react-dom": major
 ---
 
-feat(react-dom): is prefix, ms suffix for InView
+feat(react-dom): is prefix, for InView
--- a/packages/react-dom/src/useInView.spec.tsx
+++ b/packages/react-dom/src/useInView.spec.tsx
@@ -11,7 +11,7 @@ const HookComponent = ({ options, unmount }: { options?: InViewOptions; unmount?
   const wrapper = useInView(options)
   return (
     <div data-testid="wrapper" ref={!unmount ? wrapper.ref : undefined}>
-      {wrapper.inView.toString()}
+      {wrapper.isInView.toString()}
     </div>
   )
 }
@@ -26,7 +26,7 @@ const LazyHookComponent = ({ options }: { options?: InViewOptions }) => {
   if (isLoading) return <div>Loading</div>
   return (
     <div data-testid="wrapper" ref={wrapper.ref}>
-      {wrapper.inView.toString()}
+      {wrapper.isInView.toString()}
     </div>
   )
 }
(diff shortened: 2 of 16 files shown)


gwansikk (commented):
Is it necessary to explicitly use `ms` as a suffix? I'm curious about the decision to include it.  
JavaScript Web APIs (e.g., `setTimeout`) and TanStack Query's `staleTime` also use milliseconds, yet they don't include `ms` as a suffix.

CC. @kangju2000 I'm curious what you think about this too.

manudeli:
> Just wondering — is the reason for changing inView to isInView to make it clearer that it’s a boolean? I noticed that [react-intersection-observer](https://github.com/thebuilder/react-intersection-observer?tab=readme-ov-file&rgh-link-date=2025-04-17T16%3A25%3A22Z#options) uses inView as a boolean, so I was curious if there's a specific reason behind the rename.

Because we use is prefix for below case, simple reason
It's last time to make BREAKING CHANGE because @suspensive/react's version is yet 0

```tsx
import { Delay } from '@suspensive/react'
 
export const Example = () => {
  return (
    <Delay ms={200}>
      {({ isDelayed }) => (
        <div
          style={{
            opacity: isDelayed ? 1 : 0,
            transition: 'opacity 1500ms',
          }}
        ></div>
      )}
    </Delay>
  )
}
```

manudeli:
> Is it necessary to explicitly use `ms` as a suffix? I'm curious about the decision to include it. JavaScript Web APIs (e.g., `setTimeout`) and TanStack Query's `staleTime` also use milliseconds, yet they don't include `ms` as a suffix.
> 
> CC. @kangju2000 I'm curious what you think about this too.

In my opinion, delayMs is not must. I'll revert it!

kangju2000:
@gwansikk 
I usually read delay as milliseconds, so delay feels natural to me.
But if we end up needing a seconds-based prop later on, adding Ms here could help with clarity.

.changeset/six-impalas-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@suspensive/react-dom": major
+---
+
+feat(react-dom): is prefix, ms suffix for InView
```
manudeli, line 5:
```suggestion
feat(react-dom): is prefix, for InView
```

packages/react-dom/src/useInView.spec.tsx
```diff
@@ -104,16 +104,16 @@ it('should trigger onChange', () => {
...
 
   mockAllIsIntersecting(true)
-  expect(onChange).toHaveBeenLastCalledWith(
-    true,
-    expect.objectContaining({ intersectionRatio: 1, isIntersecting: true })
-  )
+  expect(onChange).toHaveBeenLastCalledWith({
+    isInView: true,
+    entry: expect.objectContaining({ intersectionRatio: 1, isIntersecting: true }),
+  })
```
manudeli, line 110: This is changed too @kangju2000 @gwansikk Check it out please

manudeli:
@gwansikk @kangju2000 I revert delayMs to delay Thanks to check that!

manudeli:
@gwansikk @kangju2000 I added you guys as co-authors in [2f3bce8](https://github.com/toss/suspensive/pull/1528/commits/2f3bce82baaf590e472ce6726f5ca02a50c92420)

packages/react-dom/src/useInView.spec.tsx
kangju2000: Makes sense to me. No strong opinion either way.

kangju2000 (approved):
🚀🚀

gwansikk (approved):
I agree to the change. go go!

Issue #1539, opened 2025-04-23
[Feature]: suspensive v3 migration codemod


gwansikk:
### Package Scope

@suspensive/codemods

### Description

- `@suspensive/react` breaking changes codemods
- `@suspensive/react-query` breaking changes codemods

### Possible Solution

No response

### etc.

No response

PR #1551, opened 2025-05-03
feat(codemods): add migrate with api codemod

gwansikk:
# Overview

related: #1539 

Add a codemod that converts the `wrap` API to `with`(`ErrorBoundaryGroup`, `ErrorBoundary`, `Suspense`) for the breaking change in `@suspensive/react v3`

- Remove `wrap` from the import declaration and declare only the APIs being used.
- Convert the HoC components used in `wrap` to the `with` interface.

## PoC

image

- This codemod has been applied and confirmed to work properly by the enterprise team I belong to.
- Out of a total of 551 components, 60 components using `wrap` were reliably converted in a 28-second operation.

## AS-IS
import { wrap } from '@suspensive/react'
import { useSuspenseQuery } from '@suspensive/react-query'

```jsx
export const Case1 = wrap
  .ErrorBoundaryGroup({ blockOutside: false })
  .ErrorBoundary({
    fallback: ({ error }) => <>{error.message}</>,
    onError: logger.log,
  })
  .Suspense({ fallback: <>loading...</>, clientOnly: true })
  .on(() => {
    const query = useSuspenseQuery({
      queryKey: ['key'],
      queryFn: () => api.text(),
    })
    return <>{query.data.text}</>
  })
```

## TO-BE
import { ErrorBoundaryGroup, ErrorBoundary, Suspense } from '@suspensive/react';
import { useSuspenseQuery } from '@suspensive/react-query'

```jsx
export const Case1 = ErrorBoundaryGroup.with({ blockOutside: false }, ErrorBoundary.with({
    fallback: ({ error }) => <>{error.message}</>,
    onError: logger.log,
  }, Suspense.with({ fallback: <>loading...</>, clientOnly: true }, () => {
    const query = useSuspenseQuery({
      queryKey: ['key'],
      queryFn: () => api.text(),
    })
    return <>{query.data.text}</>
  })))
```

## PR Checklist

3. I added documents and tests.

### DIFF
--- a/examples/react-native-playground/package.json
+++ b/examples/react-native-playground/package.json
@@ -26,6 +26,7 @@
     "react-native-web": "catalog:react19"
   },
   "devDependencies": {
+    "@suspensive/eslint-config": "workspace:*",
     "@babel/core": "^7.27.1",
     "@types/jest": "^29.5.14",
     "@types/react": "catalog:react19",
--- a/.changeset/lucky-pugs-type.md
+++ b/.changeset/lucky-pugs-type.md
@@ -1,0 +1,5 @@
+---
+"@suspensive/codemods": patch
+---
+
+feat(codemods): add migrate with api codemod
--- a/docs/suspensive.org/src/content/en/docs/codemods/_meta.tsx
+++ b/docs/suspensive.org/src/content/en/docs/codemods/_meta.tsx
@@ -8,4 +8,8 @@ export default {
     title: 'Codemods',
   },
   tanstackQueryImport: { title: 'Migrate TanStack Query imports' },
+  migrateQueryClientConsumerProps: {
+    title: 'Migrate <QueryClientConsumer/> Props',
+  },
+  migrateWithAPI: { title: 'Migrate with API' },
 } satisfies MetaRecord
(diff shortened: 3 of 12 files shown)


examples/react-native-playground/package.json
```diff
@@ -26,6 +26,7 @@
     "react-native-web": "catalog:react19"
   },
   "devDependencies": {
+    "@suspensive/eslint-config": "workspace:*",
```
gwansikk, line 29: Before pnpm version 10, some packages like eslint or prettier were automatically hoisted to the root and recognized in subprojects without separate installation.

However, starting from version 10, no packages are hoisted to the root by default, so you need to explicitly declare them in the subprojects or add hoist settings for things to work properly.

ps: In the next tasks, I will check the PNPM 10 breaking changes to improve the environment.

ref:

* [https://github.com/pnpm/pnpm/issues/8378](https://github.com/pnpm/pnpm/issues/8378)
* [https://github.com/orgs/pnpm/discussions/8945](https://github.com/orgs/pnpm/discussions/8945)

manudeli (approved):
Cool 🚀 Let's release

examples/react-native-playground/package.json
manudeli: So much thanks!!

PR #1562, opened 2025-05-09
fix(docs): align sandpack template dependencies with package.json

### LINKED ISSUE #1560: [Question]: Version mismatch between package.json and baseTemplate.ts
## Package Scope
- [x] "@suspensive/react"
- [x] "@suspensive/react-dom" 
- [x] "@suspensive/react-query"
- [ ] "@suspensive/react-native"
- [ ] "@suspensive/react-query-4"
- [ ] "@suspensive/react-query-5"
- [ ] "@suspensive/jotai"
- [ ] "@suspensive/codemods"
- [ ] etc

## Description
I noticed a potential version mismatch between dependencies defined in `docs/suspensive.org/package.json` and the ones used in Sandpack examples in `docs/suspensive.org/src/components/Sandpack/baseTemplate.ts`

In `package.json`

```json
"dependencies": {
  "@suspensive/react": "workspace:*",
  "@suspensive/react-query-4": "workspace:*",
  // other dependencies...
}
```

But in baseTemplate.ts

```json
"dependencies": {
  "@tanstack/react-query": "^4",
  "@suspensive/react": "^3.0.0-next",
  "@suspensive/react-dom": "^3.0.0-next",
  "@suspensive/react-query": "^3.0.0-next",
},
```

I wanted to understand

Is this intentional?
Does workspace:* in a monorepo context work differently from explicit versioning in Sandpack examples?
Should these be aligned, and if so, what's the recommended approach?

## Possible Solution
If these should be aligned, I see a few options:

Update baseTemplate.ts to use the same workspace references
Use a consistent explicit version in both files
Perhaps there's another approach I'm not aware of within monorepo context

I'm not familiar enough with the project's structure to determine the best approach, so would appreciate guidance on how to handle this.

Here's how I came up with the fix

Image

jiji-hoon96:
# Overview

This PR addresses the version mismatch between the dependencies defined in `package.json` and those used in Sandpack examples in `baseTemplate.ts`. (Fixes #1560)

- Updated TanStack React Query version to be more specific (`^4.36.1` instead of just `^4`)
- Changed Suspensive package references to use workspace syntax (`workspace:*`) to match how they're defined in package.json
- This ensures that the Sandpack examples use the same package versions as the rest of the project

## PR Checklist


docs/suspensive.org/src/components/Sandpack/baseTemplate.ts
```diff
@@ -24,10 +24,10 @@ export default function App() {
...
   },
   dependencies: {
-    '@tanstack/react-query': '^4',
-    '@suspensive/react': '^3.0.0-next',
-    '@suspensive/react-dom': '^3.0.0-next',
-    '@suspensive/react-query': '^3.0.0-next',
+    '@tanstack/react-query': '^4.36.1',
+    '@suspensive/react': 'workspace:*',
+    '@suspensive/react-dom': 'workspace:*',
+    '@suspensive/react-query-4': 'workspace:*',
```
manudeli, line 30:
```suggestion
    '@tanstack/react-query': '^4',
    '@suspensive/react': '^3',
    '@suspensive/react-dom': '^3',
    '@suspensive/react-query': '^3',
```

manudeli (commented):
We should get version dynamically

jiji-hoon96:
Understood.
I've changed the code dynamically, thanks for suggesting a fix

manudeli (approved):
Thanks!

PR #1574, opened 2025-05-14
fix(react-query): replace wildcard exports with named exports

### LINKED ISSUE #1563: [Bug]:It's currently unsupported to use "export *" in a client boundary with Next.js 15.2.4
### Package Scope

@suspensive/react-query

## Description
When using `@suspensive/react-query` with Next.js 15.2.4, the following error occurs during build time:

Image

Error: It's currently unsupported to use "export *" in a client boundary. Please use named exports instead.

## Error Details
The error occurs in the client boundary because of the wildcard exports used in the package:
https://github.com/toss/suspensive/blob/main/packages/react-query/src/index.ts

The error is specifically pointing to:
```js
./node_modules/.pnpm/@suspensive+react-query@3.1.1_@tanstack+react-query@5.68.0_react@19.0.0__react@19.0.0/node_modules/@suspensive/react-query/dist/index.js
```

### Expected behavior

_No response_

### To Reproduce

1. Create a Next.js 15.2.4 project
2. Install `@suspensive/react-query`
3. Use [SuspenseQuery](http://_vscodecontentref_/0) component in a React Server Component that renders a Client Component

### Possible Solution

I noticed that `@suspensive/react-query` already has version 4 available. I'm wondering if:

1. Should we use a version-specific import like `@suspensive/react-query-5` to resolve this issue? 
2. Or do you have plans to change the barrel file structure to be compatible with Next.js 15.2.4's client boundary limitations?

Since Next.js 15.2.4 now explicitly rejects wildcard exports in client boundaries, it would be helpful to know the recommended approach for users facing this issue.

### etc.

_No response_

gwansikk:
# Overview

closed: #1563

Tried to solve it with an Tsup(ESBuild) plugin or a build script, but it requires more research, so I’ll first resolve the wildcard export issue using the simplest method.

## PR Checklist


gwansikk:
| Before | After |
| --- | --- |
|image|image|

Fundamentally, TanStack Query cannot be used in server components. (At least for now\...) However, the reason this export patch is necessary is because it displays an appropriate error message to the user. I believe it's right for Next.js to show a proper message to the user.

What do you think about this? @manudeli @kangju2000 @rldnd

manudeli:
image

@gwansikk This issue is occured because of render prop prohibited in next.js's server component.
This PR makes next.js error work correctly.

manudeli (approved):
Thanks 👍

Issue #1534, opened 2025-04-19
[Feature]: useClientLayoutEffect instead of useIsomorphicLayoutEffect


manudeli:
### Package Scope

@suspensive/react, @suspensive/react-dom

### Description

We need to add @jungpaeng as co-author

https://github.com/toss/react-simplikit/issues/188

### Possible Solution

```ts
export const useClientLayoutEffect = typeof document === 'undefined' ? noop : useLayoutEffect;

```

PR #1608, opened 2025-06-04
test(react/useIsClient): migrate from 'useIsomorphicLayoutEffect' to 'useClientLayoutEffect'

sukvvon:
# Overview

* relate #1534 

## PR Checklist

### DIFF
--- a/packages/react/src/hooks/useIsClient.spec.tsx
+++ b/packages/react/src/hooks/useIsClient.spec.tsx
@@ -1,30 +1,34 @@
 import { renderHook } from '@testing-library/react'
-import { useEffect, useLayoutEffect, useState } from 'react'
+import { useLayoutEffect, useState } from 'react'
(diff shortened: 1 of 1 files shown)


packages/react/src/hooks/useIsClient.spec.tsx
```diff
@@ -51,7 +55,7 @@ describe('useIsClient', () => {
     expect(mockUseIsClient).toBeCalledTimes(2)
   })
   it('improve legacy useIsClientOnly', () => {
-    const { useIsomorphicLayoutEffect } = importUseIsomorphicLayoutEffect()
+    const { useClientLayoutEffect } = importUseClientLayoutEffect()
```
manudeli, line 58: Our legacy code was useIsomorphicLayoutEffect right. so if we should compare correctly. we need to keep this

PR #1610, opened 2025-06-05
test(jotai/SetAtom): add test for error when calling setAtom on read-only atom

sukvvon:
# Overview

## PR Checklist

### DIFF
--- a/packages/jotai/src/SetAtom.spec.tsx
+++ b/packages/jotai/src/SetAtom.spec.tsx
@@ -1,8 +1,9 @@
 import { act, fireEvent, render, screen } from '@testing-library/react'
-import { atom } from 'jotai'
+import { type WritableAtom, atom } from 'jotai'
 import { Suspense } from 'react'
 import { AtomValue } from './AtomValue'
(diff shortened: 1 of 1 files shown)


packages/jotai/src/SetAtom.spec.tsx
```diff
@@ -65,6 +66,29 @@ describe('<SetAtom />', () => {
...
+              Set 246
+            </button>
+          )
+        }}
+      </SetAtom>
+    )
+
+    expect(screen.getByText('Set 246')).toBeInTheDocument()
+  })
+
```
manudeli, line 91: In my opinion, this is not Suspensive's spec; this is Jotai's spec

sukvvon: > In my opinion, this is not Suspensive's spec; this is Jotai's spec

@manudeli 

The intention was to verify whether Suspensive’s `SetAtom` fully performed the Jotai’s built-in features

manudeli: I knew your intention, but I think that it's not our responsibility

Issue #1464, opened 2025-03-04
100% coverage


manudeli:
### Package Scope

etc

### Description

We should test all

manudeli:
@sukvvon If you have interested in suspensive contribution, could you do this please?

sukvvon:
> [@sukvvon](https://github.com/sukvvon) If you have interested in suspensive contribution, could you do this please?

@manudeli Thanks for your suggestion, i’m not sure how long it will take, but let’s try it step by step.

PR #1612, opened 2025-06-09
test(react-query-5/Mutation): add test for data update and reset

sukvvon:
# Overview

* relate #1464 

## before

image

## after

image

## PR Checklist

### DIFF
--- a/packages/react-query-5/src/Mutation.spec.tsx
+++ b/packages/react-query-5/src/Mutation.spec.tsx
@@ -1,0 +1,51 @@
+import { MutationCache, QueryCache, QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { fireEvent, render, screen } from '@testing-library/react'
+import type { ReactNode } from 'react'
+import { Mutation } from './Mutation'
+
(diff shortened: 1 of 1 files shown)


manudeli:
Sorry, this is not within our scope of tests. Could you please add tests for @suspensive/react-query instead of this? That would be very helpful for us!

sukvvon:
> Sorry, this is not within our scope of tests. Could you please add tests for @suspensive/react-query instead of this? That would be very helpful for us!

@manudeli 

Then how about removing `react-query-5` and `react-query-4` from `component_management` in `codecov.yml`?

manudeli:
> > Sorry, this is not within our scope of tests. Could you please add tests for @suspensive/react-query instead of this? That would be very helpful for us!
> 
> @manudeli
> 
> Then how about removing `react-query-5` and `react-query-4` from `component_management` in `codecov.yml`?

I think it would be better to leave it for testing in our own implementation.

For example, the networkMode: 'always' in useSuspenseQuery of "@suspensive/react-query-v4" is also should be tested.
We can reproduce to test why we set it to networkMode: 'always'.

Issue #1448, opened 2025-03-01
[Feature]: use fake timer to improve test involved with timer


manudeli:
### Package Scope

etc

### Description

ref: https://github.com/TanStack/query/pull/8720

PR #1624, opened 2025-06-21
test(react/Suspense): reduce overall test time with 'useFakeTimers' and 'useRealTimers'

sukvvon:
# Overview

* relate #1448 

## before

image

## after

image

## PR Checklist

### DIFF
--- a/packages/react/src/Suspense.spec.tsx
+++ b/packages/react/src/Suspense.spec.tsx
@@ -5,7 +5,12 @@ import { Suspense } from './Suspense'
 import { FALLBACK, Suspend, TEXT } from './test-utils'
 
 describe('<Suspense/>', () => {
-  beforeEach(() => Suspend.reset())
+  beforeEach(() => vi.useFakeTimers())
+
+  afterEach(() => {
+    vi.useRealTimers()
+    Suspend.reset()
+  })
 
   it('should render the children if nothing to suspend', () => {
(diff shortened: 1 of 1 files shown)


packages/react/src/Suspense.spec.tsx
```diff
@@ -19,18 +20,25 @@ describe('<Suspense/>', () => {
         <Suspend during={Infinity} toShow={TEXT} />
       </Suspense>
     )
+
     expect(screen.queryByText(FALLBACK)).toBeInTheDocument()
     expect(screen.queryByText(TEXT)).not.toBeInTheDocument()
   })
 
   it('should render the children after suspending', async () => {
+    vi.useFakeTimers()
```
manudeli: Move this with beforeEach on describe

packages/react/src/Suspense.spec.tsx
```diff
@@ -63,6 +78,7 @@ describe('<Suspense clientOnly/>', () => {
         {TEXT}
       </Suspense>
     )
+
```
manudeli, line 83: No unnecessary change

sukvvon: @manudeli 

I believe this is necessary to maintain consistency with other test cases and improve readability.

manudeli: I merge this change, but I think this is not improving readability.

manudeli (approved):

PR #1636, opened 2025-06-24
test(react-dom/InView): rename mock function from 'callback' to 'onChange'

sukvvon:
# Overview

## PR Checklist


manudeli:
I think `callback` is more appropriate.

sukvvon:
> Sorry, I think `callback` is more appropriate.

@manudeli 

https://github.com/toss/suspensive/blob/main/packages/react-dom/src/useInView.spec.tsx#L104

I think it would be better to align it with the `onChange` in `useInview.spec.ts` for consistency.

PR #1651, opened 2025-07-10
feat(react): add `lazy`

marshallku:
# Overview

Add logics for lazy load module safely

## PR Checklist

image

I also did check the test is passing.

## What we can do further

- Make it configurable (like adding `maxRetry` option)
- Remove dependency to `window.sessionStorage`

Closes #1651

### DIFF
--- a/packages/react/src/lazy.ts
+++ b/packages/react/src/lazy.ts
@@ -142,12 +142,16 @@ interface ReloadOnErrorOptions extends LazyOptions {
   retryDelay?: number | ((retryCount: number) => number)
   /**
    * The storage to use for the retry count. \
-   * If not provided, it assumes that you are in a browser environment and uses `sessionStorage`.
+   * If not provided, it assumes that you are in a browser environment and uses `window.sessionStorage`.
+   *
+   * @default window.sessionStorage
    */
   storage?: ReloadOnErrorStorage
   /**
    * The function to use to reload the component. \
    * If not provided, it assumes that you are in a browser environment and uses `window.location.reload`.
+   *
+   * @default window.location.reload
    */
   reload?: (timeoutId: ReturnType<typeof setTimeout>) => void
 }
@@ -172,52 +176,16 @@ export const reloadOnError = ({
   reload,
   ...options
 }: ReloadOnErrorOptions): LazyOptions => {
-  const getDefaultStorage = (): ReloadOnErrorStorage => {
-    if (storage) {
-      return storage
-    }
-
-    if (typeof window !== 'undefined' && 'sessionStorage' in window) {
-      return window.sessionStorage
-    }
-
+  const reloadStorage = (() => {
+    if (storage) return storage
+    if (typeof window !== 'undefined' && 'sessionStorage' in window) return window.sessionStorage
     throw new Error('[@suspensive/react] No storage provided and no sessionStorage in window')
-  }
-
-  const getDefaultReloadFunction = (): NonNullable<ReloadOnErrorOptions['reload']> => {
-    if (reload) {
-      return reload
-    }
-
-    if (typeof window !== 'undefined' && 'location' in window) {
-      return () => window.location.reload()
-    }
-
+  })()
+  const reloadFunction = (() => {
+    if (reload) return reload
+    if (typeof window !== 'undefined' && 'location' in window) return () => window.location.reload()
     throw new Error('[@suspensive/react] No reload function provided and no location in window')
-  }
-
-  const reloadStorage = getDefaultStorage()
-  const reloadFunction = getDefaultReloadFunction()
-
-  /**
-   * Parses the reload count from the storage.
-   */
-  const parseReload = (value: string | null): { reloadCount: number; isNaN: boolean } => {
-    if (!value) {
-      return {
-        reloadCount: 0,
-        isNaN: true,
-      }
-    }
-
-    const reloadCount = parseInt(value, 10)
-    const isNaN = Number.isNaN(reloadCount)
-
-    return {
-      reloadCount: isNaN ? 0 : reloadCount,
-      isNaN,
-    }
-  }
+  })()
 
   return {
     ...options,
@@ -233,24 +201,16 @@ export const reloadOnError = ({
 
       if (typeof retry === 'number') {
         const storedValue = reloadStorage.getItem(storageKey)
-        const parsed = parseReload(storedValue)
-        currentRetryCount = parsed.reloadCount
-
-        if (parsed.isNaN) {
-          reloadStorage.removeItem(storageKey)
+        if (storedValue) {
+          const reloadCount = parseInt(storedValue, 10)
+          if (Number.isNaN(reloadCount)) reloadStorage.removeItem(storageKey)
+          currentRetryCount = reloadCount
         }
       }
 
       const shouldRetry = retry === true || (typeof retry === 'number' && currentRetryCount < retry)
-
-      if (!shouldRetry) {
-        return
-      }
-
-      if (typeof retry === 'number') {
-        reloadStorage.setItem(storageKey, String(currentRetryCount + 1))
-      }
-
+      if (!shouldRetry) return
+      if (typeof retry === 'number') reloadStorage.setItem(storageKey, String(currentRetryCount + 1))
       const delayValue = typeof retryDelay === 'function' ? retryDelay(currentRetryCount) : retryDelay
       const timeoutId = setTimeout(() => {
         reloadFunction(timeoutId)


manudeli:
@marshallku @gwansikk @kangju2000 Let's make PR to this branch to improve this feature

.changeset/shy-planets-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@suspensive/react": minor
+---
+
+feat(react): add `safeLazy`
```
gwansikk:
```suggestion
feat(react): add `lazy`
```

packages/react/src/lazy.ts
```diff
@@ -0,0 +1,119 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const lazy = <T extends ComponentType<any>>(
+    load: () => Promise<{ default: T }>,
+    options?: { reload: number }
+  ): LazyExoticComponent<T> & {
+    load: () => Promise<void>
+  } => {
+    const finalOptions = options || defaultOptions
+
+    if (process.env.NODE_ENV === 'development' && finalOptions.reload) {
```
kangju2000: I thought `finalOptions.reload` is always set. Is there a reason for this if condition?

manudeli:
I reflect @gwansikk's [this comment](https://github.com/toss/suspensive/pull/1652#discussion_r2201322403) about using Our Callout component in [f90c549](https://github.com/toss/suspensive/pull/1651/commits/f90c5496a9b0182c6499a26599392041d5577f37)

image

examples/visualization/src/app/react/lazy/page.tsx
```diff
@@ -0,0 +1,30 @@
+'use client'
+
+import { Suspense, lazy } from '@suspensive/react'
+import React, { useState } from 'react'
+
+const Comp = lazy(() => import('./Comp'), { reload: 3 })
```
manudeli:
```suggestion
const Comp = lazy(() => import('./Comp'), { onError: () => window.location.reload() })
```

I was worried because I was not sure if the option name reload would be easy to understand, but what if I handle it like this in onError? I think it could be used in various ways. I think I could even induce a refresh by displaying a toast.

manudeli: `@suspensive/react` should also support `react-native`, so this one, which doesn't use sessionStorage internally, seems like a good general-purpose option.

marshallku: I was thinking the exact same way!
I planned to do this - removing dependency to browser - after #1656

manudeli: cc. @gwansikk @kangju2000

gwansikk: I agree. I think it's appropriate for the user to take action when an error occurs.

gwansikk: But if we remove `reload` and use `onError`, how can we implement retrying `window.location.reload()`? (Wouldn't it get a bit complicated?)

I agree with creating an escape hatch using `onError`, but if implementing `retry` is complicated, I think a `retry` option is also necessary.

manudeli: I think we can convert like below

### Infinite reload

```tsx

const customLazy = lazy.create({ reload: true })

const customLazy = lazy.create({ onError: () => window.location.reload() })
```

### Reload only at specific times.

Below is pseudo code.

```tsx
const MAX_RELOAD = 2

const customLazy = lazy.create({ reload: MAX_RELOAD })

const customLazy = lazy.create({ 
  onSuccess:  ({ load }) => {
    const storageKey = load.toString()
    window.sessionStorage.removeItem(storageKey)
  },
  onError: ({ load }) => {
    const storageKey = load.toString()
    const sessionStorageReloadedCount = parseInt(window.sessionStorage.getItem(storageKey), 10)
    const reloadedCount = Number.isNaN(sessionStorageReloadedCount) ? 0 : sessionStorageReloadedCount    
    window.sessionStorage.setItem(storageKey, String(reloadedCount + 1))
    if(reloadedCount <= MAX_RELOAD) window.location.reload()
  }
})
```

manudeli: I made Pull Request to replace `reload` with `onSuccess` and `onError` https://github.com/toss/suspensive/pull/1658

marshallku:
> @marshallku Could you make `reloadOnError`(tentative name) to support what we discussed ( https://github.com/toss/suspensive/pull/1658#issuecomment-3072636005  )?

Sure!

manudeli:
I have similar concerns, but I think that the `clientOnly` prop of Suspensive Suspense and interfaces of TanStack Query are similar cases with this lazy.

In situations where SSR is possible, the best implementation would be to do SSR without the clientOnly option, but because there are many practical difficulties, clientOnly seems to be used well. This clientOnly can also be implemented directly by users, but suspensive provides it because it deals with a realistic suspense environment.

It would be nice if we could use a cool way to network with server like graphql without TanStack Query, but realistically, since the existing server cannot be changed at all, I think it is used as a good alternative interface. TanStack Query can also be implemented directly by users, but many developers rely on TanStack Query to get the best case alternatively.

There is no right answer to how far to expose an interface, but if `lazy` with `reloadOnError` is supported in suspensive, I would probably use this because it is tested code rather than implementing it myself.

However, I hope that `reloadOnError` is treated as an alternative to the version skew solution, not as a best case for Suspensive lazy usage.

So, rather than providing `sessionStorage` and `reload` as properties of `lazyOptions`, I thought that providing reloadOnError as a separate function would be better, even though it is more complicated to use.

Conclusion: I don't know either..! But I think I'll use it if I have it. We have BC ❤️

packages/react/src/lazy.ts
```diff
@@ -0,0 +1,260 @@
...
+  } => {
+    const composedOnSuccess = ({ load }: { load: () => Promise<void> }) => {
+      defaultOptions.onSuccess?.({ load })
+      options?.onSuccess?.({ load })
+    }
+
+    const composedOnError = ({ error, load }: { error: unknown; load: () => Promise<void> }) => {
+      defaultOptions.onError?.({ error, load })
+      options?.onError?.({ error, load })
+    }
```
manudeli: Order should be considered

.changeset/shy-planets-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@suspensive/react": minor
+---
+
+feat(react): add `lazy`
```
manudeli:
```suggestion
feat(react): add `lazy`, `reloadOnError`
```

packages/react/src/lazy.ts
gwansikk: First of all, there is something I previously explained incorrectly. In TanStack Query, the options of `useMutation` (such as `onSuccess`, etc.) are triggered first, and then the mutate options (`onSuccess`) are triggered second.

ref: https://tanstack.com/query/latest/docs/framework/react/guides/mutations#mutation-side-effects

```ts
useMutation({
  mutationFn: addTodo,
  onError: (error, variables, context) => {
    // I will fire first
  },
})

mutate(todo, {
  onError: (error, variables, context) => {
    // I will fire second!
  },
})
```

gwansikk: The order of handler execution is more appropriate as `options` → `defaultOptions`.

### 1. Lazy loading is a technical process, not user interaction

Since lazy loading is a technical process involving asynchronous chunk loading, it's more appropriate to execute the handler defined at the usage site (`options`) first, and then run the global handler (`defaultOptions`) afterwards.

→ `defaultOptions` serves as the final fallback (e.g., `window.location.reload()`).

### 2. It allows for a staged flow.

If the load succeeds, UI handling (e.g., alert) can be prioritized in `options`, followed by post-processing such as logging or global state updates in `defaultOptions`.

If a forced termination occurs in `options`, `defaultOptions` will be skipped.
For example, if `window.location.reload()` is triggered in `options.onError`, the subsequent logic is interrupted.

```ts
const customLazy = lazy.create({
  onError: ({ error }) => {
    window.location.reload() // Force reload: the default onError will not be executed afterward
  },
})

const LazyComponent = customLazy(() => import('./NonExistentComponent'), {
  onError: ({ error }) => {
    logger('lazy on error:', error) // not trigger
  },
})
```

marshallku: I'm fully convinced @gwansikk!
In this case, default handler must fire at the end.

manudeli:
We can consider [jitter](https://github.com/TanStack/query/discussions/4490) to implement retry logic

packages/react/src/lazy.ts
```diff
@@ -0,0 +1,260 @@
...
+  create: createLazy,
+})
+
+interface ReloadOnErrorStorage {
+  getItem: (key: string) => string | null
+  setItem: (key: string, value: string) => void
+  removeItem: (key: string) => void
+}
+
+interface ReloadOnErrorOptions extends Omit<LazyOptions, 'preventMerging'> {
```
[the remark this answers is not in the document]
manudeli: @marshallku I think we don't need to Omit preventMerging! there is no preventMerging!
And I recommend to use OmitKeyof instead of Omit directly. OmitKeyof prevent unnecessary omitting key of any object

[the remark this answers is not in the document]
manudeli: Like this if you used OmitKeyof instead of Omit

image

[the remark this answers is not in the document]
manudeli: I removed Omit myself

manudeli (approved):
I love this new feature❤️ It's experimental feature so let's improve if we need

manudeli:
@gwansikk @kangju2000 could you review or approve this pr please

gwansikk (approved):
I love it! gogo!

PR #1652, opened 2025-07-10
fix(react): re-struct lazy interface

manudeli:


packages/react/src/lazy.ts
```diff
@@ -0,0 +1,43 @@
...
+                window.sessionStorage.setItem(storageKey, 'true')
+                window.location.reload()
+                return { default: (() => null) as unknown as T }
+              }
+              throw error
+            }
+          }
+        : load
+    ),
+    { load: () => load().then(noop) }
```
marshallku: Can you explain me about what `noop` does? I can see that it's just void function, just curious of what it does actually.

packages/react/src/lazy.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @example
+ * ```tsx
+ * import { lazy } from '@suspensive/react'
+ *
+ * const Component = lazy(() => import('./Component'), { safe: true })
+ * ```
+ */
+export const lazy = <T extends ComponentType<any>>(
+  load: () => Promise<{ default: T }>,
+  options?: { safe: boolean }
```
marshallku:
```suggestion
  options?: { safe?: boolean }
```

IMO, this option is not required for every cases.

packages/react/src/lazy.ts
```diff
@@ -0,0 +1,43 @@
...
+): LazyExoticComponent<T> & {
+  load: () => Promise<void>
+} =>
+  Object.assign(
+    originalLazy(
+      options?.safe
```
marshallku:
```suggestion
  options: { safe: boolean } = {}
): LazyExoticComponent<T> & {
  load: () => Promise<void>
} =>
  Object.assign(
    originalLazy(
      options.safe
```

Adding default value can reduce optional chainings

packages/react/src/lazy.spec.tsx
```diff
@@ -78,15 +79,20 @@ describe('default `safeLazy` function', () => {
...
     }).rejects.toThrow('Failed to load component')
 
     expect(importFunction).toHaveBeenCalledTimes(2)
     expect(reloadFunction).toHaveBeenCalledTimes(1)
   })
+  it('should reload the page only once if the dd fails to load', () => {
+    const importFunction = () => sleep(10).then(() => ({ default: () => 'hihihih' }))
+    const Component = lazy(importFunction, { safe: true })
+    expect(() => act(() => render(<Component />))).not.toThrow()
+  })
```
marshallku: Nice success case!

marshallku (commented):
Thanks for your feedback!

packages/react/src/lazy.ts
gwansikk:
```suggestion
  options?: { safe: boolean } // `reload` or `retry`
```
Is the expression "safe" appropriate? What does it mean to "safe" something—fallback or load? Can users easily anticipate the behavior? Is there a clearer expression? I'd appreciate your thoughts.

@marshallku @manudeli @kangju2000

marshallku: Nice suggestion! I'm trying to add `maxRetry` option, so name `retry` seems more appropriate.

gwansikk: That's a good! It might be a good approach to design the option interface inspired by TanStack Query Retries.

ref: [https://tanstack.com/query/v5/docs/framework/react/guides/query-retries](https://tanstack.com/query/v5/docs/framework/react/guides/query-retries)

manudeli:
Good refs

1. https://github.com/gregberge/loadable-components 
2. https://github.com/silverhand-io/react-safe-lazy

packages/react/src/lazy.ts
manudeli: I thought we should prevent below case

```tsx
lazy(() => import('...'), {}) // empty object
```

I wanna permit only below cases

```tsx
lazy(() => import('...'))
lazy(() => import('...'), { reload: 5 })
```

packages/react/src/lazy.spec.tsx
manudeli: I added more test cases
But I'm worried because I'm not sure if this test is correct. It's not easy to test import.

packages/react/src/lazy.ts
manudeli: I update this option to be `reload`: number because we can alias this implementation as 'retry' instead of our default implementation.

image

I think we can add the `retry` option like this. ref: https://blog.logto.io/react-safe-lazy/?ref=github

packages/react/src/lazy.ts
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * // Using lazy.create for factory pattern
+ * const customLazy = lazy.create({ reload: 2 })
+ * const CustomComponent = customLazy(() => import('./CustomComponent'))
+ * ```
+ *
+ * @returns A lazy component with additional `load` method for preloading
+ * @property {() => Promise<void>} load - Preloads the component without rendering it. Useful for prefetching components in the background.
+ */
+export const lazy = Object.assign(createLazy({ reload: 3 }), {
```
manudeli: I'm really debating whether it would be better to set the default reload to 0 or 3.

1. If it's 0, it seems like it would be much easier to migrate from React's lazy.
2. If it's 3, you can use this feature out of the box without having to learn any additional options.

```suggestion
export const lazy = Object.assign(createLazy({ reload: 0 }), {
```

packages/react/src/lazy.spec.tsx
```diff
@@ -0,0 +1,438 @@
...
+  }
+  key() {
+    return this.items.keys().next().value ?? null
+  }
+}
+
+const storage = new MockStorage()
+const mockReload = vi.fn<() => void>()
+
+describe('lazy', () => {
```
manudeli, line 98: I wanna get strict review for these test codes 🥲

packages/react/src/lazy.ts
gwansikk: I think case 1 is appropriate.

The `lazy` API is the same as the original `lazy` (`React.lazy`). If you don't look at the import statement, it could be misleading. I think if the default value is set to 3, users might not notice, so I believe it's better to make it explicit.

packages/react/src/lazy.ts
manudeli: I wanna prevent using Awaited<Promise<{ default: T }>>

Like this

```tsx
const Comp = lazy(() => import("./Comp"))

// Prevent using x
const x = await Comp.load()
```

manudeli:
> Good refs
> 
> 1. https://github.com/gregberge/loadable-components
> 2. https://github.com/silverhand-io/react-safe-lazy

https://github.com/jamiebuilds/react-loadable

Good ref more

docs/suspensive.org/src/content/ko/docs/react/lazy.mdx
```diff
@@ -0,0 +1,115 @@
+# lazy
+
+> **⚠️ 실험적**: 이것은 매우 실험적인 기능입니다. 프로덕션 환경에서 사용할 때 주의하세요.
+
+`lazy` 함수는 컴포넌트 로딩이 실패할 때 자동 오류 복구를 제공하는 React의 `lazy` 함수의 안전한 래퍼입니다. 컴포넌트를 로드하려고 시도하고, 로딩이 실패하면 설정된 횟수만큼 페이지를 새로고침하여 네트워크 문제나 일시적인 로딩 실패 시 더 나은 사용자 경험을 제공합니다.
```
gwansikk, line 5:
```suggestion
import { Callout } from 'nextra/components'

# lazy

<Callout type='experimental'>

이것은 매우 실험적인 기능입니다. 프로덕션 환경에서 사용할 때 주의하세요.

</Callout>

`lazy` 함수는 컴포넌트 로딩이 실패할 때 자동 오류 복구를 제공하는 React의 `lazy` 함수의 안전한 래퍼입니다. 컴포넌트를 로드하려고 시도하고, 로딩이 실패하면 설정된 횟수만큼 페이지를 새로고침하여 네트워크 문제나 일시적인 로딩 실패 시 더 나은 사용자 경험을 제공합니다.
```

packages/react/src/lazy.ts
```diff
@@ -0,0 +1,118 @@
...
+    load: () => Promise<void>
+  } => {
+    const finalOptions = options || defaultOptions
+
+    if (process.env.NODE_ENV === 'development' && finalOptions.reload) {
+      if (!Number.isInteger(finalOptions.reload) || finalOptions.reload <= 0) {
+        console.error(
+          `[@suspensive/react] lazy: reload option must be a positive integer, but received ${finalOptions.reload}. ` +
+            `Please provide a positive integer value (e.g., { reload: 3 }).`
+        )
```
gwansikk, line 38: 👍👍👍

packages/react/src/lazy.spec.tsx
gwansikk: I reviewed it and there doesn't seem to be any problem. Since it's an experimental version, I think it's okay to release it.

gwansikk (approved):
I love it!! gogo!

packages/react/src/lazy.ts
manudeli: I reflect your suggestion. I update this default reload as 1
I think it is enough as your opinion 👍

manudeli:
Good discussion Thanks!! ❤️

PR #1656, opened 2025-07-14
feat(react): Add options for reloading component infinitely

marshallku:
# Overview

- Add option for infinite reloading
- Refactor codes for improving readability

## PR Checklist


packages/react/src/lazy.ts
```diff
@@ -1,6 +1,89 @@
 import { type ComponentType, type LazyExoticComponent, lazy as originalLazy } from 'react'
 import { noop } from './utils/noop'
 
+interface ReloadOptions {
+  reload: number | true
+}
```
manudeli: This option can gradually include other options, so I think this is `LazyOptions` because it's for the lazy function, not only for reload. One of the options is `reload`. I think `onSuccess` option or `onError` option could be added in this LazyOptions

```suggestion
interface LazyOptions {
  reload: number | true
}
```

packages/react/src/lazy.ts
```diff
@@ -1,6 +1,89 @@
...
+
+const DEFAULT_RELOAD_COUNT = 1
```
manudeli, line 8: @marshallku @gwansikk @kangju2000 

I thought a lot about how much this `DEFAULT_RELOAD_COUNT` should be. To sum it up, I thought it should be 0.

First, I think this option should be used when infrastructure-based solutions are lacking, and it is not a perfect solution to the bundle skewing problem, but rather a complementary part. I am cautious about forcing the use of this reload option.

If this option is required by default for library users, it seems possible to create a lazy that reloads by default according to the user's environment through `createLazy`.

`@suspensive/react` provides various options for `Suspense`, but it is designed to allow drop-in replacement from vanilla react to `@suspensive/react`. Therefore, I thought it would be better to guarantee that lazy behaves the same in order to have no problems with immediate migration. I thought it would be good to introduce this method as an alternative to infrastructure-based solutions.

```suggestion
const DEFAULT_RELOAD_COUNT = 0
```

I wanna get your thoughts about my opinion!

packages/react/src/lazy.ts
```diff
@@ -84,12 +130,15 @@ const createLazy = (defaultOptions: { reload: number }) => {
  * ```tsx
  * import { lazy, Suspense } from '@suspensive/react'
  *
- * // Basic usage with default reload (3 times)
+ * // Basic usage with default reload (1 time)
```
manudeli, line 140:
```suggestion
 * // Basic usage is same with original react
```

docs/suspensive.org/src/content/en/docs/react/lazy.mdx
```diff
@@ -69,6 +71,24 @@ function App() {
 }
 ```
 
+#### Infinite Reload
+
+```tsx
+import { lazy, Suspense, ErrorBoundary } from '@suspensive/react'
+
+const UserProfile = lazy(() => import('./UserProfile'), { reload: true })
```
manudeli, line 85: `true` is easy to use IMO, cool!

packages/react/src/lazy.spec.tsx
```diff
@@ -142,6 +142,23 @@ describe('lazy', () => {
     expect(mockReload).toHaveBeenCalledTimes(1)
   })
 
+  it('should reload infinitely if reload is true', async () => {
```
manudeli, line 159: Easy to use 👍

packages/react/src/lazy.ts
```diff
@@ -21,55 +104,18 @@ import { noop } from './utils/noop'
...
-
-    if (process.env.NODE_ENV === 'development' && finalOptions.reload) {
-      if (!Number.isInteger(finalOptions.reload) || finalOptions.reload <= 0) {
-        console.error(
-          `[@suspensive/react] lazy: reload option must be a positive integer, but received ${finalOptions.reload}. ` +
-            `Please provide a positive integer value (e.g., { reload: 3 }).`
-        )
-      }
-    }
+    const finalOptions = parseOptions(options || defaultOptions)
```
manudeli: I'd like to follow TanStack's naming, it's really trivial things. I wanna share good refs

https://github.com/TanStack/query/blob/35d5d55aa72a9e39284554852983b0745deecc52/packages/react-query/src/useBaseQuery.ts#L55

```suggestion
    const defaultedOptions = getDefaultedOptions(options, defaultOptions)
```

I think we need to get defaulted options, not just parsedOptions, because we can add more options. options could vary in near future

packages/react/src/lazy.ts
```diff
@@ -21,55 +104,18 @@ import { noop } from './utils/noop'
...
+    const loader = finalOptions.reload ? createLoader(load, finalOptions.reload) : load
```
manudeli: I think the code will end up like this

```suggestion
    const loader = createLoader(load, defaultedOptions)
```

packages/react/src/lazy.ts
```diff
@@ -1,6 +1,89 @@
 import { type ComponentType, type LazyExoticComponent, lazy as originalLazy } from 'react'
 import { noop } from './utils/noop'
 
+interface ReloadOptions {
+  reload: number | true
```
```suggestion
  reload: number | boolean
```

IMO, reload option could disabled in this case

```tsx
const customLazy = createLazy({ reload: true })

const Comp = customLazy(() => import('...'), { reload: false }) // disable reload of defaultOptions
```

packages/react/src/lazy.ts
marshallku: I initially thought setting it to 1 would be ideal — since in my projects I often use `safeLazy` with values of `1` or `infinite`.
However, as you pointed out, defaulting it to `0` makes it a true drop-in replacement for React’s `lazy`, and I completely agree.

packages/react/src/lazy.ts
marshallku: Thought `0` can be used for that case, but `false` feels more natural.

packages/react/src/lazy.ts
gwansikk: > I thought a lot about how much this DEFAULT_RELOAD_COUNT should be. To sum it up, I thought it should be 0.

I agree as well. The API name is now `lazy`. If it's not `safeLazy`, users won't know about the reload effect. I think it's better to set the default value to 0 and make it optional.

manudeli (approved):

manudeli:
I think we can improve `lazy` more after merging this Pull Request

PR #1658, opened 2025-07-15
fix(react): replace `reload` with `onSuccess`, `onError`

manudeli:
related with: https://github.com/toss/suspensive/pull/1651#discussion_r2206136929

### DIFF
--- a/docs/suspensive.org/src/content/en/docs/react/lazy.mdx
+++ b/docs/suspensive.org/src/content/en/docs/react/lazy.mdx
@@ -192,24 +192,6 @@ const Component = lazy(() => import('./Component'), {
 })
 ```
 
-## SSR Considerations
-
-When using `lazy` with SSR frameworks like Next.js:
-
-```tsx
-import { lazy, Suspense } from '@suspensive/react'
-
-const SSRComponent = lazy(() => import('./SSRComponent'))
-
-function App() {
-  return (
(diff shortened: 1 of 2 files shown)


manudeli:
> Is it really necessary to strip out every existing behavior?
> 
> ```ts
> function createLazy({
>   onError = handleError,
>   storage,
> }: {
>   onError?: (error: Error) => void;
>   storage?: LazyStorage;   // If `window` is undefined and no `storage` is provided, this will still throw.
> } = {}) {
> }
> ```
> 
> Current implementation became very concise, but it pushes most of the core logic onto the user. As a library consumer, I — and likely many others — would probably skip Suspense altogether and roll my own solution rather than wiring up so much manually.

Or like this?

```tsx
import { lazy, reloadLazyOptions } from '@suspensive/react'

const customLazy = lazy.create(reloadLazyOptions({ storage: location.sessionStorage }))
```

I thought it is similar situation with [keepPreviousData](https://tanstack.com/query/latest/docs/framework/vue/guides/migrating-to-v5#removed-keeppreviousdata-in-favor-of-placeholderdata-identity-function)

This way can reduce the number of options and provide frequently used cases as API.

marshallku (approved):
> library should minimize BC

I’m fully convinced!
Rather than forcing users into a rigid workflow, it’s better to offer flexible, configurable options.

LGTM!

PR #1659, opened 2025-07-17
feat(react): Add option for reloading on error

marshallku:
# Overview

Add `reloadOnError` option for providing a convinent way to automatically reload when loading fails.

## PR Checklist

image

All tests are passing

### DIFF
--- a/docs/suspensive.org/src/content/en/docs/react/lazy.mdx
+++ b/docs/suspensive.org/src/content/en/docs/react/lazy.mdx
@@ -142,6 +142,142 @@ function App() {
 }
 ```
 
+### Automatic Reload on Error
+
+The `reloadOnError` function provides a convenient way to automatically reload the page when component loading fails. This is particularly useful for handling version skew issues in production environments.
+
+```tsx
+import { lazy, reloadOnError } from '@suspensive/react'
+
+// Create a lazy factory with automatic reload on error
+const customLazy = lazy.create(
+  reloadOnError({
+    retry: 3, // Retry up to 3 times
+    delay: 1000, // Wait 1 second between retries
+  })
+)
+
+const Component = customLazy(() => import('./Component'))
+```
+
+#### reloadOnError Options
+
+```tsx
+interface ReloadOnErrorOptions {
+  /**
+   * The number of times to retry the loading of the component.
+   * If `true`, the component will be retried indefinitely.
+   * @default 1
+   */
+  retry?: number | boolean
+
+  /**
+   * The delay between retries in milliseconds.
+   * If a function is provided, it will be called with the current retry count.
+   * @default 0
+   */
+  delay?: number | ((retryCount: number) => number)
+
+  /**
+   * The storage to use for tracking retry count.
+   * If not provided, uses `sessionStorage` in browser environments.
+   */
+  storage?: {
+    getItem: (key: string) => string | null
+    setItem: (key: string, value: string) => void
+    removeItem: (key: string) => void
+  }
+
+  /**
+   * The function to use to reload the component.
+   * If not provided, uses `window.location.reload()` in browser environments.
+   */
+  reload?: () => void
+
+  // Standard lazy options
+  onSuccess?: ({ load }: { load: () => Promise<void> }) => void
+  onError?: ({
+    error,
+    load,
+  }: {
+    error: unknown
+    load: () => Promise<void>
+  }) => void
+}
+```
+
+#### Advanced reloadOnError Examples
+
+**Infinite Retry with Exponential Backoff:**
+
+```tsx
+import { lazy, reloadOnError } from '@suspensive/react'
+
+const customLazy = lazy.create(
+  reloadOnError({
+    retry: true, // Infinite retries
+    delay: (retryCount) => Math.min(1000 * Math.pow(2, retryCount), 30000), // Exponential backoff, max 30s
+  })
+)
+
+const Component = customLazy(() => import('./Component'))
+```
+
+**Custom Storage and Reload Function:**
+
+```tsx
+import { lazy, reloadOnError } from '@suspensive/react'
+
+const customStorage = {
+  getItem: (key) => localStorage.getItem(key),
+  setItem: (key, value) => localStorage.setItem(key, value),
+  removeItem: (key) => localStorage.removeItem(key),
+}
+
+const customReload = () => {
+  // Custom reload logic
+  window.location.href = window.location.href
+}
+
+const customLazy = lazy.create(
+  reloadOnError({
+    retry: 5,
+    delay: 2000,
+    storage: customStorage,
+    reload: customReload,
+  })
+)
+
+const Component = customLazy(() => import('./Component'))
+```
+
+**Conditional Retry Based on Error Type:**
+
+```tsx
+import { lazy, reloadOnError } from '@suspensive/react'
+
+const customLazy = lazy.create(
+  reloadOnError({
+    retry: 3,
+    delay: 1000,
+    onError: ({ error, load }) => {
+      // Only retry for specific error types
+      if (
+        error.message?.includes('Loading chunk') ||
+        error.message?.includes('NetworkError')
+      ) {
+        console.log(
+          'Retrying due to network or chunk loading error:',
+          error.message
+        )
+      }
+    },
+  })
+)
+
+const Component = customLazy(() => import('./Component'))
+```
+
 ### Version Skew Problem Resolution
 
 ```tsx
--- a/docs/suspensive.org/src/content/en/docs/react/lazy.mdx
+++ b/docs/suspensive.org/src/content/en/docs/react/lazy.mdx
@@ -153,7 +153,7 @@ import { lazy, reloadOnError } from '@suspensive/react'
 const customLazy = lazy.create(
   reloadOnError({
     retry: 3, // Retry up to 3 times
-    delay: 1000, // Wait 1 second between retries
+    retryDelay: 1000, // Wait 1 second between retries
   })
 )
 
@@ -176,7 +176,7 @@ interface ReloadOnErrorOptions {
    * If a function is provided, it will be called with the current retry count.
    * @default 0
    */
-  delay?: number | ((retryCount: number) => number)
+  retryDelay?: number | ((retryCount: number) => number)
 
   /**
    * The storage to use for tracking retry count.
@@ -216,7 +216,7 @@ import { lazy, reloadOnError } from '@suspensive/react'
 const customLazy = lazy.create(
   reloadOnError({
     retry: true, // Infinite retries
-    delay: (retryCount) => Math.min(1000 * Math.pow(2, retryCount), 30000), // Exponential backoff, max 30s
+    retryDelay: (retryCount) => Math.min(1000 * Math.pow(2, retryCount), 30000), // Exponential backoff, max 30s
   })
 )
 
@@ -242,7 +242,7 @@ const customReload = () => {
 const customLazy = lazy.create(
   reloadOnError({
     retry: 5,
-    delay: 2000,
+    retryDelay: 2000,
     storage: customStorage,
     reload: customReload,
   })
@@ -259,7 +259,7 @@ import { lazy, reloadOnError } from '@suspensive/react'
 const customLazy = lazy.create(
   reloadOnError({
     retry: 3,
-    delay: 1000,
+    retryDelay: 1000,
     onError: ({ error, load }) => {
       // Only retry for specific error types
       if (
--- a/docs/suspensive.org/src/content/en/docs/react/lazy.mdx
+++ b/docs/suspensive.org/src/content/en/docs/react/lazy.mdx
@@ -163,7 +163,7 @@ const Component = customLazy(() => import('./Component'))
 #### reloadOnError Options
 
 ```tsx
-interface ReloadOnErrorOptions {
+interface ReloadOnErrorOptions extends LazyOptions {
   /**
    * The number of times to retry the loading of the component.
    * If `true`, the component will be retried indefinitely.
--- a/docs/suspensive.org/src/content/ko/docs/react/lazy.mdx
+++ b/docs/suspensive.org/src/content/ko/docs/react/lazy.mdx
@@ -142,9 +142,9 @@ function App() {
 }
 ```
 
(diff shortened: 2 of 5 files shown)


marshallku:
Hi @manudeli 
I’ve opened a draft PR to consolidate our ideas on the `lazy` interface.
lease feel free to share any feedback on the current implementation!

packages/react/src/lazy.ts
```diff
@@ -108,3 +109,126 @@ const createLazy =
 export const lazy = Object.assign(createLazy({}), {
   create: createLazy,
 })
+
+export interface ReloadOnErrorStorage {
```
manudeli: I thought this shouldn't be exported

```suggestion
interface ReloadOnErrorStorage {
```

packages/react/src/lazy.ts
```diff
@@ -108,3 +109,126 @@ const createLazy =
...
+  getItem: (key: string) => string | null
+  setItem: (key: string, value: string) => void
+  removeItem: (key: string) => void
+}
+
+export interface ReloadOnErrorOptions extends LazyOptions {
```
manudeli: I thought this shouldn't be exported

```suggestion
interface ReloadOnErrorOptions extends LazyOptions {
```

packages/react/src/lazy.ts
```diff
@@ -108,3 +109,126 @@ const createLazy =
...
+
+export const reloadOnError = ({ retry = 1, delay = 0, storage, reload, ...options }: ReloadOnErrorOptions) => {
+  const getDefaultStorage = (): ReloadOnErrorStorage | null => {
+    if (storage) {
+      return storage
+    }
+
+    if (typeof window === 'undefined' && 'sessionStorage' in window) {
+      return null
+    }
```
```suggestion
    if (typeof window === 'undefined' && 'sessionStorage' in window) {
      throw new Error('storage should be defined')
    }
```

packages/react/src/lazy.ts
```diff
@@ -108,3 +109,126 @@ const createLazy =
...
+  }
+
+  const getDefaultReloadFunction = (): ReloadOnErrorOptions['reload'] | null => {
+    if (reload) {
+      return reload
+    }
+
+    if (typeof window === 'undefined' && 'location' in window) {
+      return null
+    }
```
```suggestion
    if (typeof window === 'undefined' && 'location' in window) {
      throw new Error('reload should be defined')
    }
```

packages/react/src/lazy.ts
```diff
@@ -108,3 +109,126 @@ const createLazy =
...
+    onSuccess: ({ load }) => {
+      options.onSuccess?.({ load })
+      reloadStorage?.removeItem(load.toString())
+    },
+    onError: ({ error, load }) => {
+      options.onError?.({ error, load })
+
+      if (reloadStorage == null || reloadFunction == null) {
+        return
+      }
```
manudeli: reloadStroage, reloadFunction could be Nonnullable if throw error when there is no reloadStorage, reloadFunction

manudeli (commented):
It would be nice if test code was also added

docs/suspensive.org/src/content/ko/docs/react/lazy.mdx
```diff
@@ -142,7 +142,143 @@ function App() {
 }
 ```
 
-### Version Skew 문제 해결
+### 자동 오류 시 페이지 새로고침
```
gwansikk:
```suggestion
### 오류 시 자동 페이지 새로고침
```

docs/suspensive.org/src/content/ko/docs/react/lazy.mdx
```diff
@@ -142,7 +142,143 @@ function App() {
...
+
+`reloadOnError` 옵션을 사용하면 컴포넌트 로딩이 실패할 때 자동으로 페이지를 새로고침할 수 있습니다. 이는 특히 프로덕션 환경에서 버전 차이(version skew) 문제를 처리하는 데 유용합니다.
```
gwansikk:
```suggestion
`reloadOnError` 옵션을 사용하면 컴포넌트 로딩이 실패할 때 자동으로 페이지를 새로고침할 수 있습니다. 이는 특히 배포 환경에서 버전 차이(version skew) 문제를 처리하는 데 유용합니다.
```

What is a production environment? Please explain in detail. Wouldn't "deployment environment" be more appropriate?

packages/react/src/lazy.ts
```diff
@@ -108,3 +109,130 @@ const createLazy =
...
+   * @default 1
+   */
+  retry?: number | boolean
+  /**
+   * The delay between retries. \
+   * If a function is provided, it will be called with the current retry count.
+   *
+   * @default 0
+   */
+  delay?: number | ((retryCount: number) => number)
```
gwansikk:
```suggestion
  retryDelay?: number | ((retryCount: number) => number)
```

I would like the expression to be clear. (This is just a suggestion.)

related: https://tanstack.com/query/latest/docs/framework/react/guides/query-retries#retry-delay

packages/react/src/lazy.ts
```diff
@@ -108,3 +109,130 @@ const createLazy =
...
+      }
+
+      if (typeof retry === 'number') {
+        reloadStorage.setItem(storageKey, String(currentRetryCount + 1))
+      }
+
+      const delayValue = typeof delay === 'function' ? delay(currentRetryCount) : delay
+      setTimeout(() => {
+        reloadFunction()
+      }, delayValue)
```
gwansikk: This structure uses `setTimeout`, and it seems like there could be a potential for memory leaks if the `reloadFunction` is overridden with a user-defined function instead of using the default (`window.location.reload`).

It's not an explicit or guaranteed issue, but there's a possibility of situations where closures or timers are repeatedly registered or not properly cleaned up. Wouldn't it be a good idea to consider handling this with something like `clearTimeout`? (e.g. in `onSuccess`)

I'm a bit concerned because the structure doesn't guarantee that `reloadFunction` will always trigger a window reload. What do you think?

CC. @manudeli @marshallku

gwansikk (commented):
let's go go! so cool 😎

packages/react/src/lazy.ts
marshallku: Thought retrying is just only thing to delay, but making it accurate could be even better. Thanks!

packages/react/src/lazy.ts
marshallku: I assumed that reloading the page would automatically clear all timers and reset the JavaScript execution context. \
However, since `reloadFunction` can contain custom logic, preventing memory leaks becomes crucial. \

I’ve updated the code to pass `timerId` into `reloadFunction` to let users manage and clear their timeouts explicitly.

packages/react/src/lazy.ts
```diff
@@ -108,3 +109,130 @@ const createLazy =
...
+   */
+  storage?: ReloadOnErrorStorage
+  /**
+   * The function to use to reload the component. \
+   * If not provided, it assumes that you are in a browser environment and uses `window.location.reload`.
+   */
+  reload?: (timeoutId: ReturnType<typeof setTimeout>) => void
+}
+
+export const reloadOnError = ({
```
manudeli, line 168:
```suggestion
/** @experimental This is experimental feature. */
export const reloadOnError = ({
```

marshallku:
> @marshallku We need to export reloadOnError in index.ts to make it public API

Thanks @manudeli! I exported it

docs/suspensive.org/src/content/ko/docs/react/lazy.mdx
```diff
@@ -142,7 +142,143 @@ function App() {
...
+  })
+)
+
+const Component = customLazy(() => import('./Component'))
+```
+
+#### reloadOnError 옵션
+
+```tsx
+interface ReloadOnErrorOptions {
```
manudeli:
```suggestion
interface ReloadOnErrorOptions extends LazyOptions {
```

docs/suspensive.org/src/content/en/docs/react/lazy.mdx
```diff
@@ -142,6 +142,142 @@ function App() {
...
+  })
+)
+
+const Component = customLazy(() => import('./Component'))
+```
+
+#### reloadOnError Options
+
+```tsx
+interface ReloadOnErrorOptions {
```
manudeli:
```suggestion
interface ReloadOnErrorOptions extends LazyOptions {
```

packages/react/src/index.ts
```diff
@@ -4,7 +4,7 @@ export { ErrorBoundary, useErrorBoundary, useErrorBoundaryFallbackProps } from '
 export { ErrorBoundaryGroup, useErrorBoundaryGroup } from './ErrorBoundaryGroup'
 export { Delay } from './Delay'
 export { ClientOnly } from './ClientOnly'
-export { lazy } from './lazy'
+export { lazy, reloadOnError } from './lazy'
```
manudeli, line 7: Let's go 🚀

packages/react/src/lazy.spec.tsx
```diff
@@ -285,4 +282,112 @@ describe('lazy', () => {
       expect(onError).toHaveBeenCalledWith({ error: expect.any(Error), load: expect.any(Function) })
     })
   })
+
+  describe('reloadOnError option', () => {
```
manudeli, line 273: Can we define specs for below case?

```tsx

const customLazy = lazy.create(reloadOnError({ ... }))
const Comp = customLazy(() => import('...'), { onSuccess, onError }) // I expect this don't destroy reloadOnError
```

```tsx

const Comp = lazy(() => import('...'), reloadOnError({ onSuccess, onError })) // I expect these onSuccess, onError don't destroy reloadOnError
```

```tsx

const customLazy = lazy.create(reloadOnError({ onSuccess, onError })) // I expect these onSuccess, onError don't destroy reloadOnError
const Comp = customLazy(() => import('...')) 
```

I think this is my last feedback for this feature

marshallku:
```ts

const customLazy = lazy.create(reloadOnError({ ... }))
const Comp = customLazy(() => import('...'), { onSuccess, onError }) // I expect this don't destroy reloadOnError
```

Hey @manudeli, I expect this case should destroy `reloadOnError`, as we're testing `it should merge default options with provided options`. \
Maybe we might add private flag like `_preventMerging` to achieve this.

I agree with second and third cases!

manudeli: As @marshallku said, in this case, since what reloadOnError returns is just overriding the onSuccess and onError passed when calling lazy.create, it seems natural to destroy the reloadOnError function.

Then, what if we remove the override and allow the onSuccess and onError passed in to the customLazy generated from lazy.create to coexist without overriding the default onSuccess and onError?

```tsx

const customLazy = lazy.create(reloadOnError({ onSuccess, onError: () => overlay.open(() => <Toast />)}))
const Comp = customLazy(() => import('...'), { onSuccess, onError: () => console.error('load error')}) // this don't destroy reloadOnError, onSuccess, onError

const Example = () => {
  return (
    <Suspense fallback="loading...">
      <Comp /> {/* When we fail load splitted code */}
    </Suspense>
  )
}

// 1. fail load splitted code
// 2. console.error('load error')
// 3. overlay.open(() => <Toast />)
// 4. location.reload()
```

@marshallku I think it would be more easy to use because we don't have to worry about overriding it unexpectedly. Sorry for the constant changes to the interface decisions. There are some things I didn't think about. @gwansikk @kangju2000 I think this part is an important decision. If you have a different opinion about my direction, please feel free to suggest a different direction 🙏

marshallku: I’ve added the `preventMerging` option in commit 9b54bb0a06b7c24c4303fd9c823b9ac8efd94fa4 (and tests in 7edf64a938ad1200287ec6cf0a5d6a7697f1cb5d). \
My only concern is that this might introduce too many options for users.

Overriding `onSuccess` feels unnatural, but without the `preventMerging` option, users can’t disable the default behavior without creating a new lazy instance.

Please feel free to giving me feedbacks, @manudeli @gwansikk @kangju2000

manudeli: I think it would be good idea if preventMerging is default behavior. and remove preventMerging option! could we?

> users can’t disable the default behavior without creating a new lazy instance.

If we can't disable default behavior, I think that is easy to understand than before

I think our situation similar with below codes

```tsx
const { mutate } = useMutation({ 
  onSuccess: () => {}
})

mutate(,{
  onSuccess: () => {} // This don't disable onSuccess callback passed in useMuatation hook, and it's still easy
})


const customLazy = lazy.create({ onSuccess: () => {} })
const Comp = customLazy(() => import('...'), { onSuccess: () => {} })

Comp.load({
  onSuccess: () => {} // I think we can make this too..! 😁
})
```

marshallku: Makes sense to me!

manudeli: Thank you for the long discussion! I think we finally found the best direction.

packages/react/src/lazy.ts
```diff
@@ -36,23 +36,33 @@ interface LazyOptions {
...
-    const defaultedOptions = { ...defaultOptions, ...options }
+    const composedOnSuccess = ({ load }: { load: () => Promise<void> }) => {
+      defaultOptions.onSuccess?.({ load })
+      options?.onSuccess?.({ load })
+    }
+
+    const composedOnError = ({ error, load }: { error: unknown; load: () => Promise<void> }) => {
+      defaultOptions.onError?.({ error, load })
+      options?.onError?.({ error, load })
+    }
```
manudeli, line 54: I don't think this order is wrong, but I'm not sure if it's right. It would be nice to come up with a reason for the order.

marshallku:
I think it's ready to be merged too!

packages/react/src/lazy.ts
gwansikk: I also agree with @manudeli . @tanstack/query follows the order of options -> defaultOptions. (Of course, this doesn't mean we have to follow it, but I personally think this order feels natural.)

It might be good to consider which order would be better.

manudeli: > Of course, this doesn't mean we have to follow it, but I personally think this order feels natural.

My opinion is same

manudeli:
Merge first improve detail after! Let's merge this

PR #1663, opened 2025-07-20
fix(react): enable additional props and add 'name' prop to Suspense component

manudeli:


packages/react/src/Suspense.tsx
```diff
@@ -16,16 +16,13 @@ export interface SuspenseProps extends ReactSuspenseProps {
  * @see {@link https://suspensive.org/docs/react/Suspense Suspensive Docs}
  */
 export const Suspense = Object.assign(
-  ({ clientOnly, children, fallback }: SuspenseProps) => {
+  ({ clientOnly, fallback, ...rest }: SuspenseProps) => {
```
manudeli: name could be contained in rest object because React.SuspenseProps have name prop

```tsx
interface SuspenseProps {
  children?: ReactNode | undefined;

  /** A fallback react tree to show when a Suspense child (like React.lazy) suspends */
  fallback?: ReactNode;

  /**
  * A name for this Suspense boundary for instrumentation purposes.
  * The name will help identify this boundary in React DevTools.
  */
  name?: string | undefined;
}
```

packages/react/src/Suspense.tsx
```diff
@@ -16,15 +16,16 @@ export interface SuspenseProps extends ReactSuspenseProps {
  * @see {@link https://suspensive.org/docs/react/Suspense Suspensive Docs}
  */
 export const Suspense = Object.assign(
-  ({ clientOnly, children, fallback }: SuspenseProps) => {
+  ({ clientOnly, fallback, children, name, ...rest }: SuspenseProps) => {
```
manudeli, line 19: name should be passed because React.SuspenseProps have name prop

```tsx
interface SuspenseProps {
  children?: ReactNode | undefined;

  /** A fallback react tree to show when a Suspense child (like React.lazy) suspends */
  fallback?: ReactNode;

  /**
  * A name for this Suspense boundary for instrumentation purposes.
  * The name will help identify this boundary in React DevTools.
  */
  name?: string | undefined;
}
```

gwansikk (approved):
LGTM! Thanks you!

manudeli:
> How about adding test code?
> 
> e.g.
> 
> ```ts
>   it('should set Component.everything', () => {
>     const Component = ErrorBoundary.with({ fallback: () => <></> }, () => <></>)
>     Component.everything = () => {}
> 
>     expect(Component.everything).toBeDefined()
>   })
> ```

I tried to make it possible, but I failed.. Could you try it please?

gwansikk:
The type issue with `additional props` still persists. Fundamentally, type extension (merging) does not work when using `Object.assign` due to its characteristics. I’m looking for a way to resolve this,

but is there a specific reason for using `Object.assign`?

@manudeli

PR #1670, opened 2025-07-28
refactor(react): enhance reloadOnErrorOptions by clarifying default storage and reload function documentation

manudeli:


packages/react/src/lazy.ts
```diff
@@ -142,12 +142,16 @@ interface ReloadOnErrorOptions extends LazyOptions {
...
+   * If not provided, it assumes that you are in a browser environment and uses `window.sessionStorage`.
+   *
+   * @default window.sessionStorage
    */
   storage?: ReloadOnErrorStorage
   /**
    * The function to use to reload the component. \
    * If not provided, it assumes that you are in a browser environment and uses `window.location.reload`.
+   *
+   * @default window.location.reload
```
manudeli, line 154: improved jsdoc

packages/react/src/lazy.ts
```diff
@@ -172,52 +176,16 @@ export const reloadOnError = ({
...
-
-    const reloadCount = parseInt(value, 10)
-    const isNaN = Number.isNaN(reloadCount)
-
-    return {
-      reloadCount: isNaN ? 0 : reloadCount,
-      isNaN,
-    }
-  }
+  })()
```
manudeli, line 188: It's trivial, iife could remove unnecessary naming

packages/react/src/lazy.ts
```diff
@@ -172,52 +176,16 @@ export const reloadOnError = ({
...
-    }
-
-    const reloadCount = parseInt(value, 10)
-    const isNaN = Number.isNaN(reloadCount)
-
-    return {
-      reloadCount: isNaN ? 0 : reloadCount,
-      isNaN,
-    }
-  }
```
manudeli, line 220: I think this could be removed to reduce the library bundle size.

manudeli: image

marshallku: Oh, I thought it must be minified but not

manudeli: I think the `minify` option was a good attempt! While there may be some disagreement, I believe distributing libraries without minification provides a better DX experience and allows for optional minification from the library user's perspective. Thank you for giving me another chance to reconsider.

manudeli: @marshallku If you agree with this change, please approve it.

PR #1671, opened 2025-07-28
chore(tsup): Do minify build output

marshallku:
Screenshot 2025-07-29 at 08 41 48
Screenshot 2025-07-29 at 08 40 55

It drops down size of build output.

### DIFF
--- a/configs/tsup/src/index.ts
+++ b/configs/tsup/src/index.ts
@@ -8,6 +8,7 @@ export const options: Options = {
   outDir: 'dist',
   sourcemap: true,
   dts: true,
+  minify: true,
 }
 
 export const scriptOptions: Options = {


manudeli:
image

`minify` reduce size but also reduce readability too for library users, and minifying could be excuted in service using suspensive 🙏🏻

marshallku:
> `minify` reduce size but also reduce readability too for library users, and minifying could be excuted in service using suspensive. so I think not using `minify` is better for more readable library 🙏🏻

Do users actually read build output? \
I often dive into the source code of the libraries I use - such as bunch of npm packages, or even like cpython, and swc too - but I never tried to read build output if it's open source project.

marshallku:
After checking a few libraries, I realized that minified output isn’t universal. \
I’d assumed build outputs were like binary files—unreadable and optimized purely for performance—but npm packages don’t follow that pattern.

manudeli:
> > `minify` reduce size but also reduce readability too for library users, and minifying could be excuted in service using suspensive. so I think not using `minify` is better for more readable library 🙏🏻
> 
> Do users actually read build output? I often dive into the source code of the libraries I use - such as bunch of npm packages, or even like cpython, and swc too - but I never tried to read build output if it's open source project.

This may not be the case for everyone, but at least I read the built files and quickly determine the exact error cause. Furthermore, if minification is necessary, many frameworks handle it in their service code by default, so I think it's better not to minify libraries for the sake of development productivity.

manudeli:
Still, it was a good attempt, and it gave me a chance to explore this option again! Thank you.

PR #1680, opened 2025-08-09
docs(suspensive.org): add Marshall Ku to maintainers list

### LINKED ISSUE #1650: [Feature]: Add safeLazy for safely load components dynamically
### Package Scope

@suspensive/react

### Description

Instead of using `lazy` in react, I want to load module safely without runtime panics.

### Possible Solution

If the an error occurs while loading module, reloading the page may resolve the problem.

### etc.

_No response_

manudeli:
@marshallku Welcome to suspensive 🎉

### TO BE

image


marshallku (approved):
Thank you for welcoming me!!

PR #1683, opened 2025-08-10
test(react/DefaultProps): fix unused 'clientOnly3' variable and simplify by reusing single variable

sukvvon:
# Overview

## PR Checklist


manudeli:
> Hmmm, if we simplify this, what would be the benefit? (Convince me) 🤔
> 
> I think the previous approach is easier to distinguish and makes it simpler to understand the context.

I agree with your idea. I couldn't understand why this needs improvement. An explanation is needed

packages/react/src/DefaultProps.spec.tsx
```diff
@@ -94,41 +94,40 @@ describe('<DefaultPropsProvider/>', () => {
...
         {createElement(() => {
-          clientOnly2 = useContext(SuspenseDefaultPropsContext).clientOnly
+          clientOnly = useContext(SuspenseDefaultPropsContext).clientOnly
           return <></>
         })}
       </DefaultPropsProvider>
     )
 
-    expect(clientOnly3).toBeUndefined()
+    expect(clientOnly).toBeUndefined()
```
sukvvon, line 130: @manudeli @gwansikk 

I noticed that in the original code, `clientOnly3` was declared but the context was actually being assigned to `clientOnly2`, leaving `clientOnly3` unused. This meant that `expect(clientOnly3).toBeUndefined()` would always pass without actually testing the intended behavior, since `clientOnly3` was never reassigned. Rather than just fixing the variable names, I simplified the approach to ensure we're actually verifying the intended behavior.

manudeli: I understand! Thanks for the explanation!

PR #1684, opened 2025-08-11
refactor(react): replace `lazy.create` by `createLazy`

manudeli:
## AS-IS lazy.create
```tsx
// src/utils/lazy.ts
export const customLazy = lazy.create({})

// src/components/Comp.tsx
- import { lazy } from 'react'
+ import { customLazy } from 'src/utils/lazy' // We can't drop-in replacement without rename
const Comp = customLazy(() => import())
```

## TO-BE createLazy

```tsx
// src/utils/lazy.ts
import { createLazy } from '@suspensive/react'
export const lazy = createLazy({})

// src/components/Comp.tsx
- import { lazy } from 'react'
+ import { lazy } from 'src/utils/lazy'  // We can drop-in replacement
const Comp = lazy(() => import()) // Same with React.lazy
```

## PR Checklist

### DIFF
--- a/packages/react/src/lazy.spec.tsx
+++ b/packages/react/src/lazy.spec.tsx
@@ -272,13 +272,13 @@ describe('lazy', () => {
     it('should execute component onError first, then default onError', async () => {
       const mockImport = importCache.createImport({ failureCount: 10, failureDelay: 100, successDelay: 100 })
       const callOrder: string[] = []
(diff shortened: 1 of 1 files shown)


gwansikk (approved):
I agree to change `lazy.create` to `createLazy`.

This naturally creates an environment where the previous code base (`React.lazy`) can be easily modified.

PR #1693, opened 2025-08-17
test(react/lazy): comprehensive test improvements and fix 'ImportCache.isCached' to use 'null' check instead of 'undefined'

sukvvon:
# Overview

- Fix `ImportCache.isCached()` to use strict `null` check (`!== null`) instead of `undefined` check
- Refactor preload test with improved state assertions and cleaner event handlers
- Replace fragment with `div` element for consistent DOM structure
- Enhance caching test logic with distinct component paths for better isolation
- Optimize timing values across all tests for more reliable execution
- Add micro-timing adjustments with precise state transition verification
- Improve error handling tests with better element selection using `getAllByText`
- Add explicit button type attributes for enhanced accessibility
- Strengthen test stability with more granular timing controls and assertions

## PR Checklist

### DIFF
--- a/packages/react-native/jest.config.ts
+++ b/packages/react-native/jest.config.ts
@@ -7,4 +7,5 @@ module.exports = {
   },
   transformIgnorePatterns: ['../../node_modules/(?!@react-native|react-native)'],
   setupFilesAfterEnv: ['./jest.setup.ts'],
+  collectCoverage: true,
 } satisfies Config
--- a/packages/react-native/package.json
+++ b/packages/react-native/package.json
@@ -43,11 +43,11 @@
     "ci:attw": "attw --pack",
     "ci:eslint": "eslint \"**/*.{ts,tsx,cts,mts}\"",
     "ci:publint": "publint --strict",
-    "ci:test": "jest --coverage",
+    "ci:test": "jest",
     "ci:type": "tsc --noEmit",
     "clean": "rimraf ./dist ./coverage ./node_modules",
     "prepack": "pnpm run build",
-    "test:ui": "jest --watchAll --coverage"
+    "test:ui": "jest --watchAll"
   },
   "devDependencies": {
     "@suspensive/eslint-config": "workspace:*",


sukvvon (commented):
IMO, we need `isLoaded` in `PathData`

packages/react/src/lazy.spec.tsx
```diff
@@ -11,59 +11,73 @@ type PathData = {
...
   }
+
   getAttempt(path: string): number {
     return this.pathData.get(path)?.attempt || 0
   }
+
   isCached(path: string): boolean {
-    return this.pathData.get(path)?.cache !== undefined
+    return this.pathData.get(path)?.cache !== null
   }
```
sukvvon, line 79: The `isCached()` method was checking `!== undefined` but the cache field is typed as `Promise | null` and initialized as `null`. So i replaced `undefined` with `null`.

packages/react/src/lazy.spec.tsx
```diff
@@ -118,53 +138,65 @@ describe('lazy', () => {
...
+      expect(importCache.getAttempt('/cached-component2')).toBe(1)
+      expect(importCache.isCached('/cached-component2')).toBe(false)
+      await act(() => vi.advanceTimersByTimeAsync(100))
+      expect(importCache.isCached('/cached-component2')).toBe(true)
+
+      render(<Component1 />)
+
+      expect(importCache.getAttempt('/cached-component1')).toBe(1)
+      expect(importCache.isCached('/cached-component1')).toBe(true)
     })
```
sukvvon, line 200: Then we can prove this test.

packages/react/src/lazy.spec.tsx
marshallku: So embarrassed that I missed this. 😢 

btw, @manudeli if this class is only for mocking import in tests, the cache-related logic might not be necessary since [Modules](https://www.typescriptlang.org/docs/handbook/2/modules.html) don’t have caching behavior.

PR #1705, opened 2025-08-25
docs: export Post type for usage in PostListItem component example

wo-o29:
# Overview

This PR modifies the Post type definition by adding the export keyword so that it can be imported and used in the PostListItem component example and other parts of the codebase. This improves example correctness and usage clarity.

## PR Checklist

### DIFF
--- a/docs/suspensive.org/src/content/en/docs/react-query/SuspenseInfiniteQuery.mdx
+++ b/docs/suspensive.org/src/content/en/docs/react-query/SuspenseInfiniteQuery.mdx
@@ -121,7 +121,7 @@ export const postsInfiniteQueryOptions = () =>
 ```
 
 ```tsx api.ts
-type Post = {
+export type Post = {
   userId: number
   id: number
   title: string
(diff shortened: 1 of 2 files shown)


docs/suspensive.org/src/content/en/docs/react-query/SuspenseInfiniteQuery.mdx
```diff
@@ -121,7 +121,7 @@ export const postsInfiniteQueryOptions = () =>
 ```
 
 ```tsx api.ts
-type Post = {
```
manudeli, line 124: I don't think this shouldn't be exported. why we need to export?

wo-o29: I thought `export` keyword was needed because `Post` type was imported from `./api` path in `PostListItem`!
```tsx
import type { Post } from './api' 

export const PostListItem = (props: Post) => {
  return <li style={{ marginBottom: '10px' }}>{props.title}</li>
}
```

manudeli: Ahha! I understood

PR #1716, opened 2025-08-30
ci(.github/workflows): add autofix.yml

manudeli:
I added https://autofix.ci/ to fix PRs with prettier automatically.

We can see autofix like below [db9cb51](https://github.com/toss/suspensive/pull/1716/commits/db9cb518719381cd91fe3044450d39105afab66a)

Now contributors don't have to execute `pnpm install` to format code with prettier.
I think contributors for documentation may feel easy to do something

image

autofix ci will format automatically with prettier

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@
 
 version: 2
 updates:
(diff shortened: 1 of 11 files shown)


gwansikk (approved):
I love it! 🚀

PR #1760, opened 2025-09-26
feat(react): infer ErrorBoundary error type from `shouldCatch` prop

manudeli:
# Infer ErrorBoundary error type from `shouldCatch` prop

- Updated ErrorBoundary to support automatic narrowing of error types based on provided constructors or conditions.
- Added comprehensive tests for ExtractErrorType to validate behavior with various input scenarios.
- Refactored ErrorBoundary tests to leverage new type narrowing features for better type assertions.

## AS-IS

- Passing shouldCatch does not affect types; fallback/onError receive error: Error.
- Even with arrays of constructors/type guards, error stays Error; custom fields require manual narrowing.
- Result: extra boilerplate and weaker type safety.

```tsx
class CustomError extends Error {
  customProperty = 'custom'
}

<ErrorBoundary
  shouldCatch={CustomError}
  fallback={({ error }) => {
    // ⛔ error: Error → can't access error.customProperty without manual guard
    return <div>{error.message}</div>
  }}
  onError((error) => {
    // ⛔ error: Error → can't access error.customProperty without manual guard
  })
/>

```

## TO-BE

- ErrorBoundary infers the error type from shouldCatch.
- Single constructor → error is that constructor type.
- Array of constructors/guards → error is a union of matched types.
- fallback/onError use ExtractErrorType<TShouldCatch>; fallback also receives reset.

```tsx
class CustomError extends Error {
  customProperty = 'custom'
}

<ErrorBoundary
  shouldCatch={CustomError}
  onError={(error) => { /* error: CustomError */ }}
  fallback={({ error, reset }) => {
    // ✅ error: CustomError
    error.customProperty; // OK
    return <button onClick={reset}>{error.message}</button>
  }}
  onError((error) => {
    // ✅ error: CustomError
    error.customProperty; // OK
  })
/>

```

## PR Checklist

### DIFF
--- a/packages/react/src/ErrorBoundary.tsx
+++ b/packages/react/src/ErrorBoundary.tsx
@@ -74,7 +74,7 @@ const matchError = (errorMatcher: ErrorMatcher, error: Error): error is InferErr
         return error instanceof errorMatcher
       }
     } catch {
-      // If accessing prototype throws, it's not a constructor. This can happen with Proxy objects or in restricted environments
+      // If accessing prototype throws, it's not a constructor. This can happen with proxy objects or in restricted environments.
     }
     return (errorMatcher as ErrorValidator | ErrorTypeGuard<InferError<typeof errorMatcher>>)(error)
   }


gwansikk:
It looks good to me, but I'd like to open the code myself and check if there's room for type improvements. (Please give me just one more day to review it!)

manudeli:
> It looks good to me, but I'd like to open the code myself and check if there's room for type improvements. (Please give me just one more day to review it!)

Sure!

packages/react/src/ErrorBoundary.tsx
```diff
@@ -167,28 +193,34 @@ class FallbackBoundary extends Component<{ children: ReactNode }> {
...
-          resetKeys={[group.resetKey, ...(resetKeys || [])]}
-          ref={baseErrorBoundaryRef}
-        >
-          {children}
-        </BaseErrorBoundary>
-      )
-    }
-  ),
+    return (
+      <BaseErrorBoundary<TShouldCatch>
```
gwansikk, line 232:
```suggestion
      <BaseErrorBoundary
```
Is there a reason you specified the generic here? (Is it to explicitly show it because of the type assertion below?)

gwansikk (commented):
We could reduce the type assertions further, but I'm not sure if that's really necessary right now. I think it would be fine to improve it later. It works as intended!

packages/react/src/ErrorBoundary.tsx
manudeli: I thought there was no reason not to show it. 

> Is it to explicitly show it because of the type assertion below?

That's right

packages/react/src/ErrorBoundary.tsx
```diff
@@ -59,48 +76,55 @@ export type ErrorBoundaryProps = PropsWithChildren<{
...
 
-  componentDidCatch(error: Error, info: ErrorInfo) {
+  componentDidCatch(error: InferError<TShouldCatch>, info: ErrorInfo) {
     this.props.onError?.(error, info)
   }
 
   reset = () => {
     this.props.onReset?.()
-    this.setState(initialErrorBoundaryState)
+    this.setState(initialErrorBoundaryState<InferError<TShouldCatch>>())
```
manudeli: @gwansikk I removed type assertion like this

marshallku (commented):
The type system looks somewhat complicated.
Should we actually consider cases like `shouldCatch={[CustomError, true]}`?
I’m not sure they add much value.

gwansikk (approved):
I think I'm fully prepared for the task of narrowing down the types. I believe it would be better to postpone any additional interface changes to the next version!

manudeli:
> ```tsx
> class FifthError<T> extends Error {
>   fifthProperty: string = 'fifth' as const
>   constructor(public data: T) {
>     super(`FifthError with data: ${data}`)
>   }
> }
> 
> const example = (
>     <ErrorBoundary
>       shouldCatch={FifthError<string>}
>       fallback={({ error }) => {
>         expectTypeOf(error).toEqualTypeOf<FifthError<string>>()
>         return <div>{error.message}</div>
>       }}
>     >
>       <div>content</div>
>     </ErrorBoundary>
>   )
> ```
> 
> There may not be meaningful test for this, but I think it would be good to also include the error class with generics in the tests. (It's a safeguard to avoid deviating from the original intent during future development.)

I reflected this review comment in this PR.

packages/react/src/ErrorBoundary.tsx
```diff
@@ -85,22 +122,26 @@ const initialErrorBoundaryState: ErrorBoundaryState = {
...
     const { isError } = this.state
     const { resetKeys } = this.props
     if (isError && prevState.isError && hasResetKeysChanged(prevProps.resetKeys, resetKeys)) {
       this.reset()
     }
   }
 
-  componentDidCatch(error: Error, info: ErrorInfo) {
+  componentDidCatch(error: InferError<TShouldCatch>, info: ErrorInfo) {
     this.props.onError?.(error, info)
```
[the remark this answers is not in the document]
manudeli: > The `componentDidCatch` method receives an `Error` parameter from React, but it's typed as `InferError<TShouldCatch>`. This creates a type mismatch since React's error boundary lifecycle methods always receive the base `Error` type. The parameter should remain `Error` and be cast when passed to `onError` if needed.

@copilot but componentDidCatch method could be excuted after render method. then couldn't we guarantee error is InferError<TShouldCatch>?

packages/react/src/ErrorBoundary.tsx
```diff
@@ -122,10 +171,7 @@ class BaseErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState
       if (error instanceof ErrorInFallback) {
         throw error.originalError
       }
-      const isCatch = Array.isArray(shouldCatch)
-        ? shouldCatch.some((shouldCatch) => checkErrorBoundary(shouldCatch, error))
-        : checkErrorBoundary(shouldCatch, error)
-      if (!isCatch) {
+      if (!shouldCatchError(shouldCatch, error)) {
```
manudeli, line 171: `shouldCatchError` will type-narrow to error as `InferError<typeof shouldCatch>` after this if statement.

packages/react/src/ErrorBoundary.tsx
```diff
@@ -24,30 +26,73 @@ import type { ConstructorType } from './utility-types/ConstructorType'
...
+      if (
+        errorMatcher.prototype &&
+        (errorMatcher.prototype instanceof Error || errorMatcher.prototype === Error.prototype)
+      ) {
+        return error instanceof errorMatcher
+      }
+    } catch {
+      // If accessing prototype throws, it's not a constructor
+    }
+    return (errorMatcher as ErrorValidator | ErrorTypeGuard<InferError<typeof errorMatcher>>)(error)
```
manudeli, line 79: @marshallku Isn't AS-IS enough? I wanna know your intention

marshallku: I meant to minimize the mismatch between runtime and types

packages/react/src/ErrorBoundary.tsx
```diff
@@ -24,30 +26,70 @@ import type { ConstructorType } from './utility-types/ConstructorType'
...
+
+const matchError = (errorMatcher: ErrorMatcher, error: Error): error is InferError<typeof errorMatcher> => {
+  if (typeof errorMatcher === 'boolean') {
+    return errorMatcher
   }
-  if (shouldCatch.prototype instanceof Error) {
-    return error instanceof shouldCatch
+  if (typeof errorMatcher === 'function') {
+    try {
+      if (errorMatcher.prototype instanceof Error || errorMatcher.prototype === Error.prototype) {
```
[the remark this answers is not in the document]
manudeli: I agree this suggestion, @marshallku is this better for you too?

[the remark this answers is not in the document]
marshallku: I'm with your opinion. \
I tested few cases, and it just works as I expected.

PR #1761, opened 2025-09-27
chore(deps): bump nextra to 4.5.0

SEOKKAMONI:
# Overview

https://github.com/shuding/nextra/issues/4773

At the time I added this feature, I wasn’t aware that clicking the Copy Button was copying the wrong content.
I later found out that this bug was fixed in version 4.5.0, so I updated the package and ensured that the Prompt Copy Button now works correctly.

I will continue to keep an eye on nextra updates, and contribute when there are meaningful improvements.

## PR Checklist

### DIFF
--- a/.github/workflows/broken-link-checker.yml
+++ b/.github/workflows/broken-link-checker.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   broken-link-checker:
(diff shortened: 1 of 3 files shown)


gwansikk (approved):
Go go!

PR #1763, opened 2025-09-29
feat(react): add genericForwardRef to help inter ErrorBoundary error type with using forwardRef

miinhho:
# Overview

## PR Checklist

## Summary
This pull request refactors the `ErrorBoundary` component to use a new utility, `genericForwardRef`, which allows for better support of generic props with `forwardRef` in React. The changes also include the addition of the `genericForwardRef` implementation and its type tests.

Key changes:

**Refactoring to support generic forward refs:**

* Replaced usage of `forwardRef` with the new `genericForwardRef` utility in the `ErrorBoundary` component to better support generic props and type inference.
* Updated the type signature of `ErrorBoundary` to align with the new generic forward ref approach, simplifying its export.

**New utility and tests:**

* Added a new utility, `genericForwardRef`, in `utils/genericForwardRef.tsx` to enable generic prop support with React's `forwardRef`.
* Added type-level tests for `genericForwardRef` to ensure correct type inference for props and refs.


packages/react/src/utils/genericForwardRef.tsx
```diff
@@ -0,0 +1,11 @@
+import { type PropsWithoutRef, type ReactElement, type Ref, type RefAttributes, forwardRef } from 'react'
+
+/**
+ * A utility that allows creating a forwardRef component that can accept generic props.
+ * React.forwardRef does not support this out of the box.
+ */
+export function genericForwardRef<TRef, TProp = object>(
+  render: (props: PropsWithoutRef<TProp>, ref: Ref<TRef>) => ReactElement | null
+) {
+  return forwardRef(render) as (props: TProp & RefAttributes<TRef>) => ReactElement | null
```
manudeli, line 10: If we handle it with a type assertion inside of genericForwardRef, this is just hiding problem. Furthermore, this involves adding a function to the runtime. Therefore, I think it's better to resolve it solely through TypeScript type assertion without making another function. Since this issue will likely be resolved starting with React 19, I think it's better to simply expose the as method and maintain it.

PR #1769, opened 2025-10-09
refactor(react): ErrorBoundaryState could be not `InferError<TShouldCatch>`

manudeli:


packages/react/src/ErrorBoundary.tsx
```diff
@@ -183,22 +183,24 @@ class BaseErrorBoundary<TShouldCatch extends ShouldCatch = ShouldCatch> extends
         throw error.originalError
       }
       if (!shouldCatchError<InferError<TShouldCatch>>(shouldCatch as ShouldCatch<InferError<TShouldCatch>>, error)) {
-        // eslint-disable-next-line @typescript-eslint/only-throw-error
         throw error
```
manudeli, line 189: This could be not `InferError<TShouldCatch>`. so I think `ErrorBoundaryState<InferError<TShouldCatch>>` should be just `ErrorBoundaryState` internally

And also we can remove disabling-eslint line

Issue #1444, opened 2025-02-19
[Feature]: suspense support for image


gwansikk:
### Package Scope

@suspensive/react-dom

### Description

added suspense support for image loading.

manudeli:
Can we treat this issue as v3 minor?

gwansikk:
> Can we treat this issue as v3 minor feature?

https://github.com/toss/suspensive/issues/1444#issuecomment-2673444021

manudeli:
Let's do it

gwansikk:
WIP

gwansikk:
Recently, I have been going through a busy schedule. I have not been able to proceed with the task, but I promise to complete it within this month.

PR #1774, opened 2025-10-10
feat(react): add `useSuspenseImage`, `<SuspenseImage/>`

gwansikk:
# Overview

closed: https://github.com/toss/suspensive/issues/1444

## Summary

This PR introduces a **`useSuspenseImage` hook** and a **`<SuspenseImage />` component** to the React package. They enable image loading through **React Suspense**, suspending rendering while an image is loading and resuming once it is ready. To prevent redundant work, image loading is **cached by `src`**, and for **SSR environments**, a mock image object is returned immediately so that `` tags are included in the server-rendered HTML.

## Key Point

* **Suspense-based image loading**
  Image loading throws a Promise during the pending state, allowing Suspense boundaries to control fallback rendering. Once resolved, an `HTMLImageElement` is returned.

* **Built-in caching**
  A `Map<string, Promise<HTMLImageElement>>` caches image-loading Promises by `src`, avoiding duplicate network requests and repeated instantiation.

* **SSR-friendly behavior**
  On the server, the hook does not suspend. Instead, it immediately returns a mock object shaped like `HTMLImageElement` (e.g. `{ src, complete: false }`), ensuring `` elements appear in the initial HTML output for SEO and consistency.

* **Internal `use()` utility**
  An internal helper attaches `status / value / reason` to Promises and either throws or returns based on their state, implementing a canonical Suspense pattern. (support react 18)

* **Documentation added**
  Both English and Korean documentation (`SuspenseImage.mdx`) and navigation metadata are included.

## Example

```tsx
import { Suspense, SuspenseImage, useSuspenseImage } from '@suspensive/react';

function WithComponent() {
  return (
    <Suspense fallback={<div>Loading image...</div>}>
      <SuspenseImage src="https://picsum.photos/400/300">
        {(img) => <img src={img.src} alt="example" />}
      </SuspenseImage>
    </Suspense>
  );
}

function WithHook() {
  const img = useSuspenseImage('https://picsum.photos/400/300');
  return <img src={img.src} alt="example" />;
}
```

## Result

* Consumers no longer need to manage explicit loading state for images; **a single `<Suspense>` boundary is sufficient**.
* Users can manage images declaratively. They can access elements with the guarantee that the image has been loaded.
* **Duplicate image loads are avoided** through caching, reducing unnecessary network and processing overhead.
* **SSR output consistently includes `` elements**, preventing missing images in initial HTML and improving SEO and perceived performance.

## PR Checklist


packages/react/src/AwaitImage.tsx
```diff
@@ -0,0 +1,27 @@
+import { type ReactNode, use } from 'react'
```
manudeli:
```suggestion
import { type ReactNode } from 'react'
```

`use` should be removed to support react 18
I think [this](https://github.com/dai-shi/react18-use) is good alternative

packages/react/src/SuspenseImage.tsx
```diff
@@ -0,0 +1,27 @@
...
+  })
+
+  imageCache.set(src, imageLoadPromise)
+  return imageLoadPromise
+}
+
+export function AwaitImage({ src, children }: { src: string; children: (img: HTMLImageElement) => ReactNode }) {
+  const img = use(preloadImage(src))
+  return <>{children(img)}</>
+}
```
manudeli: At first, I thought AwaitImage would be helpful.

But ultimately, I thought:
1. If I wanted to support images,
2. I also wanted to support videos,
3. I also wanted to support 3D assets.
4. etc...

I wish there was an interface that could load blobs that could handle all img, video, and 3D assets. It just loads blobs, not only img.

packages/react/src/AwaitImage.tsx
gwansikk: That's right. We need to replace the use of React19, and we are considering it. (Currently, we tried it for the PoC.)

packages/react/src/SuspenseImage.tsx
gwansikk: I agree. Let's start by implementing Image first and consider gradual expansion.

gwansikk: I agree that we need to support not only images but also other resources like videos and 3D assets,, etc.

I'd like to complete the current `SuspenseImage` implementation first, then explore this direction in a follow-up.

I'm still considering the API design approach: whether to create a single generic api that handles all resource types, or provide purpose-specific APIs for each resource type.

gwansikk:
Sorry for the delay in getting things done. I'm back now. In the new year, I’ll be able to dedicate more time to Suspensive!

packages/react/src/SuspenseImage.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+function preloadImage(src: string): Promise<HTMLImageElement> {
+  const cached = imageCache.get(src)
+  if (cached) {
+    return cached
+  }
+
+  const imageLoadPromise = new Promise<HTMLImageElement>((resolve, reject) => {
+    const img = new Image()
```
manudeli, line 58: I worry that this shouldn't work in react-native.

manudeli: @gwansikk I think you just missed this comment

gwansikk: I'm looking for an excellent way to make this work in an RN environment without using the Web DOM API. I want to abstract images solely with Promises and improve it so that it isn't affected by the platform.

gwansikk: <details>
<summary>KOR</summary>

`new Image()`는 단순한 이미지 로더가 아니라, 브라우저 렌더링 파이프라인(네트워크, 캐시, 디코딩)에 이미지를 명시적으로 등록하는 Web Platform API입니다.
이 방식을 사용하면 Suspense가 해제되는 시점이 ``가 실제로 paint 가능한 시점과 최대한 일치하게 됩니다.
즉, “Suspense가 풀린 시점 = 브라우저에서 이미지가 준비된 시점”이라는 의미적 보장을 할 수 있습니다. 이 보장은 `{ src }` 객체나 fetch 기반 접근으로는 얻기 어렵습니다.

(개인적으로는 `new Image`를 모킹한 객체를 만들어 해결하는 방식은 의미적으로 맞지 않다고 봅니다.
`new Image()`는 단순 객체 생성이 아니라, 브라우저 렌더링 파이프라인과 직접 상호작용하는 강력한 API이기 때문입니다.)

ref:
- https://html.spec.whatwg.org/multipage/images.html#decoding-images
- https://html.spec.whatwg.org/multipage/images.html#updating-the-image-data

문제는 `new Image()`가 Web 전용 API라는 점입니다. React Native에는 DOM이나 `HTMLImageElement` 개념이 없기 때문에, Web과 RN에서 동일한 구현을 공유하는 것은 구조적으로 불가능합니다. RN에서는 의미적으로 대응되는 수단이 `Image.prefetch` 정도입니다.

그래서 모든 플랫폼을 지원하려면 현실적인 선택지는 두 가지로 보입니다.

1. 라이브러리가 Web / RN 구현을 모두 제공하는 방식

```
├─ ImageView.web.tsx
├─ ImageView.native.tsx
```

* Web: `new Image()`
* RN: `Image.prefetch`
* `.web.tsx` / `.native.tsx`로 분리
* 장점: 사용자 설정 없이 바로 사용 가능, Suspense 해제 의미를 라이브러리가 일관되게 보장
* 단점: RN 구현 책임 + `react-native` peerDependency 필요
  (RN 쪽 구현을 별도 유틸/패키지로 분리하는 선택지도 가능)

2. 라이브러리는 추상 API만 제공하고, 구현은 사용자에게 위임하는 방식

```ts
// core
export interface ImageLoader {
  preload(src: string): Promise<void>;
}
```

* 라이브러리는 Promise 기반 preload 인터페이스만 정의
* Web / RN preload 구현은 사용자가 직접 제공
* 장점: 완전히 플랫폼 불가지론, 의존성 최소화
* 단점: 사용자 설정 부담, Suspense 해제 시점의 의미가 구현에 따라 달라질 수 있음

RN 지원을 하려면 라이브러리가 플랫폼별 구현을 책임질지, 아니면 구현을 사용자에게 위임할지를 고민해야할 거 같습니다.
</details>

---

`new Image()` is not just a simple image loader, but a Web Platform API that explicitly registers the image with the browser's rendering pipeline (network, cache, decoding).
Using this method allows the point at which Suspense is lifted to closely match the point when `` is actually paintable.
In other words, you can semantically guarantee that “Suspense resolution = image is ready in the browser.”
This guarantee is hard to achieve using `{ src }` objects or fetch-based approaches.

(Personally, I don't think mocking a `new Image` object is semantically correct.
`new Image()` is not just about creating an object; it's a powerful API that directly interacts with the browser’s rendering pipeline.)

ref:

* [https://html.spec.whatwg.org/multipage/images.html#decoding-images](https://html.spec.whatwg.org/multipage/images.html#decoding-images)
* [https://html.spec.whatwg.org/multipage/images.html#updating-the-image-data](https://html.spec.whatwg.org/multipage/images.html#updating-the-image-data)

The problem is that `new Image()` is a web-only API. React Native doesn’t have a DOM or the concept of `HTMLImageElement`, so it’s structurally impossible to share the same implementation between Web and RN.
The closest semantic equivalent in RN is `Image.prefetch`.

So to support all platforms, there seem to be two realistic choices:

1. The library provides both Web / RN implementations

```
├─ ImageView.web.tsx
├─ ImageView.native.tsx
```

* Web: `new Image()`
* RN: `Image.prefetch`
* Separate with `.web.tsx` / `.native.tsx`
* Pros: Works out of the box with consistent semantics around Suspense resolution
* Cons: Library must own RN implementation + needs `react-native` as a peerDependency
  (You could also extract the RN implementation into a separate utility or package)
  
2. The library provides only an abstract API, and implementation is delegated to the user

```ts
// core
export interface ImageLoader {
  preload(src: string): Promise<void>;
}
```

* The library defines a preload interface based on Promise
* Web / RN implementation is provided by the user
* Pros: Platform-agnostic, minimal dependencies
* Cons: Users bear the setup burden, and the meaning of Suspense resolution depends on user implementation

If you want to support RN, the key question is whether the library should own the platform-specific implementations or delegate them to the user.

manudeli: Because we have `@suspensive/react-dom`, I think this interface could be added in it, if we don't remove `@suspensive/react-dom`. but recently, I wanted to drop `@suspensive/react-dom` because:

1. Suspensive `<InView/>` is not better than react-intersection-observer, When I first make `<InView/>` of Suspensive, I wanna support isomorphic InView for `react-dom` and `react-native` both supported. but I couldn't do it well
2. `<FadeIn/>` could be better to give implementation to library user itself

PR #1776, opened 2025-10-11
test(react): add test for ErrorBoundary to reject empty array for shouldCatch prop

manudeli:


packages/react/src/ErrorBoundary.tsx
```diff
@@ -40,72 +40,50 @@ export interface ErrorBoundaryFallbackProps<TError extends Error = Error> extend
...
 
-/**
- * Type guard function that narrows error type
- */
-type ErrorTypeGuard<TError extends Error = Error> = (error: Error) => error is TError
-
-/**
- * Validation function without type narrowing
- */
+type ErrorTypeGuard<TError extends Error> = (error: Error) => error is TError
```
manudeli, line 43: I thought ErrorTypeGuard is already good naming. so I think we can remove jsdoc

packages/react/src/ErrorBoundary.tsx
```diff
@@ -40,72 +40,50 @@ export interface ErrorBoundaryFallbackProps<TError extends Error = Error> extend
...
 type ErrorValidator = (error: Error) => boolean
```
manudeli, line 44: I thought ErrorValidator is already good naming. so I think we can remove jsdoc

packages/react/src/ErrorBoundary.tsx
```diff
@@ -40,72 +40,50 @@ export interface ErrorBoundaryFallbackProps<TError extends Error = Error> extend
...
 
-type ErrorMatcher<TError extends Error = Error> =
-  | boolean
-  | ConstructorType<TError>
-  | ErrorTypeGuard<TError>
-  | ErrorValidator
-
-type ShouldCatch<TError extends Error = Error> = ErrorMatcher<TError> | readonly ErrorMatcher<TError>[]
+type ErrorMatcher = boolean | ConstructorType<Error> | ErrorTypeGuard<Error> | ErrorValidator
```
manudeli, line 46: IMO, ErrorMatcher shouldn't be generic

packages/react/src/ErrorBoundary.tsx
```diff
@@ -185,7 +163,11 @@ class BaseErrorBoundary<TShouldCatch extends ShouldCatch = ShouldCatch> extends
       if (error instanceof ErrorInFallback) {
         throw error.originalError
       }
-      if (!shouldCatchError<InferError<TShouldCatch>>(shouldCatch as ShouldCatch<InferError<TShouldCatch>>, error)) {
```
manudeli, line 188: I removed unnecessary type assertion and generics

PR #1826, opened 2025-11-14
fix(*): update tsdown

manudeli:


configs/tsdown/package.json
```diff
@@ -5,27 +5,10 @@
...
-        "types": "./dist/index.d.cts",
-        "default": "./dist/index.cjs"
-      }
-    },
-    "./package.json": "./package.json"
-  },
-  "main": "dist/index.cjs",
-  "module": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "main": "src/index.ts",
```
manudeli, line 8: We use only ts file for internal package, @suspensive/tsdown from now

configs/tsdown/src/index.ts
```diff
@@ -1,6 +1,6 @@
-import type { Options } from 'tsdown'
```
manudeli, line 1: There is no type Option tsdown in new version of tsdown

packages/react-query/package.json
```diff
@@ -32,8 +32,8 @@
     "./package.json": "./package.json"
   },
   "main": "dist/index.cjs",
-  "module": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "module": "dist/index.mjs",
```
manudeli, line 35: there is no js file in tsdown -> only mjs, cjs file extension from now

packages/react-query/src/bin/utils/package.ts
```diff
@@ -39,7 +39,7 @@ export function getIndexFileContent(...paths: string[]): string {
   const basePath = path.resolve(...paths, 'dist').replace(/src/, '')
 
   try {
-    return fs.readFileSync(path.join(basePath, 'index.js'), 'utf-8')
+    return fs.readFileSync(path.join(basePath, 'index.mjs'), 'utf-8')
```
manudeli, line 42: So I changed this part of @suspensive/react-query

scripts/add-use-client.js
```diff
@@ -14,7 +14,7 @@ const files = readdirSync(distDir)
 
 // Add 'use client' to ClientOnly chunk files
 const clientOnlyFiles = files.filter(
-  (file) => file.startsWith('ClientOnly-') && (file.endsWith('.js') || file.endsWith('.cjs'))
+  (file) => file.startsWith('ClientOnly-') && (file.endsWith('.js') || file.endsWith('.mjs') || file.endsWith('.cjs'))
```
manudeli, line 17: And this part

PR #1848, opened 2025-12-02
feat(react-query): remove `getQueryClient` & add `createGetQueryClient`

manudeli:
**Summary**
This PR removes the old `getQueryClient` API and introduces `createGetQueryClient`. All related tests, documentation, and examples are updated accordingly.

**Key Point**
Previously, you had to understand that **the config passed to `getQueryClient` behaved differently depending on the environment**:

* the server needed a fresh config on every request,
* the browser reused only the very first config you passed.

This meant the same code could produce different results depending on where it ran.

With `createGetQueryClient`, **the config you provide once is now consistently reused across both server and browser**.
There is no longer any difference in behavior or lifecycle that the developer must account for.

**Example**

```ts
// create once and the provided config is consistently reused everywhere
import { createGetQueryClient } from '@suspensive/react-query';

export const { getQueryClient } = createGetQueryClient({
  defaultOptions: { queries: { staleTime: 3000 } },
});
```

```ts
// server or browser – same behavior
const queryClient = getQueryClient();
```

**Result**
No more environment-specific rules.
No more “server must pass config each time” vs “browser only applies first config.”
`createGetQueryClient` ensures one definition, one config, and consistent behavior everywhere.

### DIFF
--- a/examples/next-streaming-react-query/src/app/providers.tsx
+++ b/examples/next-streaming-react-query/src/app/providers.tsx
@@ -8,8 +8,6 @@ import { getQueryClient } from './get-query-client'
 export function Providers(props: { children: ReactNode }) {
   const queryClient = getQueryClient()
 
-  console.log({ ddd: queryClient.getDefaultOptions() })
-
   return (
     <QueryClientProvider client={queryClient}>
       {props.children}


jungpaeng:
> I'm wondering which name to use between `defineGetQueryClient` and `createGetQueryClient`.

I think `createGetQueryClient` is a more appropriate name since the `defineGetQueryClient` function serves as a factory.

manudeli:
> > I'm wondering which name to use between `defineGetQueryClient` and `createGetQueryClient`.
> 
> I think `createGetQueryClient` is a more appropriate name since the `defineGetQueryClient` function serves as a factory.

@jungpaeng @gwansikk I reflect your opinion! Thanks!

docs/suspensive.org/src/content/ko/docs/react-query/createGetQueryClient.mdx
```diff
@@ -245,10 +364,39 @@ const queryClient = getQueryClient({
...
 - **서버 안전성**: SSR 환경에서 `new QueryClient()`를 직접 생성하는 대신 항상 `getQueryClient`를 사용하여 적절한 요청 격리와 OOM 방지를 보장하세요.
 
-- **브라우저 싱글톤**: 브라우저 인스턴스는 싱글톤이므로 첫 번째 호출 이후의 config 변경은 무시됩니다. 첫 번째 호출에서 QueryClient를 구성하세요.
+<Callout type='error'>
+
+**전역 변수로 사용하지 마세요**
+
+`getQueryClient`를 사용하더라도 App Router에서 `export const queryClient = getQueryClient()`처럼 전역 변수에 저장하고 다른 곳에서 import해서 사용하면 동일한 보안 위험이 발생합니다. 서버에서 QueryClient 인스턴스가 요청 간 공유되어 사용자 간 데이터 누수가 발생할 수 있습니다. 전역 변수에 저장하지 말고 필요한 곳에서 직접 `getQueryClient()`를 호출하세요.
+
+</Callout>
```
gwansikk, line 398: 👍👍

examples/next-streaming-react-query/src/app/providers.tsx
```diff
@@ -1,12 +1,15 @@
...
+
 import { QueryClientProvider } from '@tanstack/react-query'
 import { ReactQueryDevtools } from '@tanstack/react-query-devtools'
 import { type ReactNode } from 'react'
+import { getQueryClient } from './get-query-client'
 
 export function Providers(props: { children: ReactNode }) {
   const queryClient = getQueryClient()
 
+  console.log({ ddd: queryClient.getDefaultOptions() })
```
gwansikk: Is this intentional?

manudeli: oh..! not intentional

manudeli: Thanks! I removed it

PR #1857, opened 2025-12-12
refactor(react): use 'displayName' variable reference in '.with()' method

sukvvon:
# Overview

Use variable reference for `displayName` instead of hardcoded strings in `.with()` method for consistency.

**Before:**
```tsx
{ displayName: `ErrorBoundary.with(...)` }
```

**After:**
```tsx
{ displayName: `${ErrorBoundary.displayName}.with(...)` }
```

**Changed files:**
- `ErrorBoundaryGroup.tsx`
- `ErrorBoundary.tsx`
- `Suspense.tsx`
- `Delay.tsx`

(`ClientOnly.tsx` already used this pattern)

## PR Checklist


gwansikk (approved):
I can see the consistency argument if `ClientOnly.tsx` already does it this way, but I'm not convinced it's really necessary.

PR #1866, opened 2026-01-06
fix(react-query): update peer dependencies for react-query-* packages to use wildcard version

manudeli:
Issued by @preco21 @woowan

Because we intend to use @suspensive/react-query@2 with no error about peerDeps. so this should be fixed.


WooWan (approved):
👍

Issue #1688, opened 2025-08-13
[Feature]: remove unnecessary hook, queryOptions in v4


manudeli:
### Package Scope

@suspensive/react-query-4, @suspensive/react-query, @suspensive/react-query-5

### Description

lighten @suspensive/react-query in v4

PR #1890, opened 2026-01-28
feat(react-query): remove @suspensive/react-query & remove unnecessary apis of @suspensive/react-query-*

manudeli:
resolve https://github.com/toss/suspensive/issues/1688

### DIFF
--- a/docs/suspensive.org/src/content/en/docs/introduction.mdx
+++ b/docs/suspensive.org/src/content/en/docs/introduction.mdx
@@ -30,18 +30,12 @@ graph TD;
         A["@suspensive/react"]
     end
     subgraph BB ["Adapter Packages"]
-        C["@suspensive/react-query"]
+        C["@suspensive/react-query-4,5"]
         D["@suspensive/jotai"]
-    end
-    subgraph CC ["Version-Specific Packages"]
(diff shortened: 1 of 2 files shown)


packages/react-query-5/package.json
```diff
@@ -55,7 +55,7 @@
     "react": "catalog:react19"
   },
   "peerDependencies": {
-    "@tanstack/react-query": "*",
+    "@tanstack/react-query": "^5.82.0",
```
kangju2000: Is there a reason we’re pinning @tanstack/react-query to ^5.82.0 as a peer dependency?

manudeli: mutationOptions is supported by @tanstack/react-query from 5.82.0 but I think we don't have to make constraint. I'll edit it to just ^5

manudeli: I reflected your opinion in [ce53be9](https://github.com/toss/suspensive/pull/1890/commits/ce53be9b06f938a2ea3813efb4e937261c6624dd)

Thanks!

gwansikk (approved):
Go go! I believe that removing @suspensive/react-query is sustainable. Thank you!

PR #1900, opened 2026-02-02
Fix bugs in lazy and stabilize it

marshallku:
# Overview

Fixed several bugs in lazy and removed the `@experimental` tag to mark it as stable.

Additional features I'm currently working on:

- Allow lazy to return a specific component as fallback
- Implement useLazy hook

These are considered feature additions rather than stabilization work, so they will be handled in separate PRs.

## PR Checklist

### DIFF
--- a/packages/react/src/lazy.ts
+++ b/packages/react/src/lazy.ts
@@ -4,7 +4,7 @@ import { noop } from './utils/noop'
 
 interface LazyOptions {
   onSuccess?: ({ load }: { load: () => Promise<void> }) => void
-  onError?: ({ error, load }: { error: unknown; load: () => Promise<void> }) => undefined
+  onError?: ({ error, load }: { error: unknown; load: () => Promise<void> }) => void
 }
 
 /**
--- a/packages/react/src/lazy.ts
+++ b/packages/react/src/lazy.ts
@@ -210,8 +210,11 @@ export const reloadOnError = ({
         const storedValue = reloadStorage.getItem(storageKey)
         if (storedValue) {
           const reloadCount = parseInt(storedValue, 10)
-          if (Number.isNaN(reloadCount)) reloadStorage.removeItem(storageKey)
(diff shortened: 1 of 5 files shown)


packages/react/src/lazy.ts
```diff
@@ -59,7 +57,7 @@ export const createLazy =
       defaultOptions.onError?.({ error, load })
     }
 
-    const loadNoReturn = () => load().then(noop)
+    const loadNoReturn = Object.assign(() => load().then(noop), { toString: () => load.toString() })
```
marshallku: We check the loading state of a component using `Function.toString()`, but it always returns the same value when the function is set like `load().then(noop)`.

packages/react/src/lazy.ts
```diff
@@ -209,8 +203,11 @@ export const reloadOnError = ({
         const storedValue = reloadStorage.getItem(storageKey)
         if (storedValue) {
           const reloadCount = parseInt(storedValue, 10)
-          if (Number.isNaN(reloadCount)) reloadStorage.removeItem(storageKey)
-          currentRetryCount = reloadCount
+          if (Number.isNaN(reloadCount)) {
```
marshallku, line 205: `NaN` was silently killing lazy

packages/react/src/lazy.ts
```suggestion
    const loadNoReturn = load
```

Then How about this? We can just pass load directly too.

marshallku: `load` actually returns a value, so it'll break type system.

manudeli: > it'll break type system.

But it's not BREAKING CHANGE because `lazy` was an experimental interface yet. so I want you just care about which one is better implementation

manudeli: @marshallku re-alert you

marshallku: I just simplified it!

PR #1905, opened 2026-02-28
fix(react-query-4): forward filters in IsFetching

saengmotmi:
# Overview

This PR fixes `@suspensive/react-query-4`'s `<IsFetching />` so query filters are actually forwarded to TanStack Query v4.

## What changed

- Forward filter props in `<IsFetching />`:
  - `useIsFetching()` -> `useIsFetching(filter)`
- Add runtime test coverage:
  - verifies filters are passed through to `useIsFetching`
  - verifies rendered value comes from the hook result
- Keep test style aligned with existing specs:
  - partial module mocking via `vi.importActual`
  - explicit mock reset in `afterEach`

## Why

TanStack Query v4 supports `useIsFetching(filters?)`, but the current wrapper ignored incoming filters.
This made `<IsFetching queryKey={...}>` behave like unfiltered fetching count.

## Validation

- `corepack pnpm --filter @suspensive/react-query-4 exec eslint src/IsFetching.tsx src/IsFetching.spec.tsx src/IsFetching.test-d.tsx`
- `corepack pnpm --filter @suspensive/react-query-4 ci:test -- src/IsFetching.spec.tsx`

## PR Checklist


gwansikk (commented):
Thank you for your contribution! Please note that this API is currently in an experimental state and may change at any time.

The following code is the type signature of `useIsFetching`. If options (`ContextOptions`) are also added as Props, I believe it would allow usage identical to `useIsFetching`. What are your thoughts on this?

```ts
// tanstack/react-query@4
import type { ContextOptions } from './types';
import type { QueryFilters, QueryKey } from '@tanstack/query-core';
interface Options extends ContextOptions {
}
export declare function useIsFetching(filters?: QueryFilters, options?: Options): number;
export declare function useIsFetching(queryKey?: QueryKey, filters?: QueryFilters, options?: Options): number;
```

PR #1910, opened 2026-03-03
refactor(lazy): update component type definitions to require 'x' prop

manudeli:
If we stablize `lazy`, Could we think about this type narrowing? or `Promise<void>` is better for return type of `load`?


packages/react/src/lazy.ts
```diff
@@ -1,8 +1,8 @@
 import { type ComponentType, type LazyExoticComponent, lazy as originalLazy } from 'react'
 
-interface LazyOptions {
-  onSuccess?: ({ load }: { load: () => Promise<{ default: ComponentType<any> }> }) => void
-  onError?: ({ error, load }: { error: unknown; load: () => Promise<{ default: ComponentType<any> }> }) => void
+interface LazyOptions<TComponentType extends ComponentType<any>> {
+  onSuccess?: ({ load }: { load: () => Promise<{ default: TComponentType }> }) => void
+  onError?: ({ error, load }: { error: unknown; load: () => Promise<{ default: TComponentType }> }) => void
```
manudeli, line 5: @marshallku I don't know if we do type narrowing is better. but ComponentType<any> is bad so I made suggestion

PR #1911, opened 2026-03-05
feat(react): add `ErrorBoundary.Observer`

manudeli:
## new `ErrorBoundary.Observer`

`ErrorBoundary.Observer` is a component that observes errors from all nested `<ErrorBoundary/>` components. It does not catch or handle errors — it simply observes them, making it ideal for integrating with external error reporting tools like Sentry.

### Motivation

Previously, you had to attach `onError` prop individually to each `<ErrorBoundary/>`. To observe errors across the entire app, you had to repeatedly pass the same handler to every ErrorBoundary.

### AS-IS

```tsx
import { ErrorBoundary } from '@suspensive/react'
import * as Sentry from '@sentry/react'

function App() {
  return (
    <ErrorBoundary
      fallback={<FallbackA />}
      onError={Sentry.captureReactException} // repeated every time
    >
      <ErrorBoundary
        fallback={<FallbackB />}
        onError={Sentry.captureReactException} // repeated every time
      >
        <YourComponent />
      </ErrorBoundary>
    </ErrorBoundary>
  )
}
```

### TO-BE

```tsx
import { ErrorBoundary } from '@suspensive/react'
import * as Sentry from '@sentry/react'

function App() {
  return (
    <ErrorBoundary.Observer onError={Sentry.captureReactException}>
      <ErrorBoundary fallback={<FallbackA />}>
        <ErrorBoundary fallback={<FallbackB />}>
          <YourComponent />
        </ErrorBoundary>
      </ErrorBoundary>
    </ErrorBoundary.Observer>
  )
}
```

### Additive (not override)

`ErrorBoundary.Observer` does not override each `<ErrorBoundary/>`'s own `onError`. Both are always called.

```tsx
import { ErrorBoundary } from '@suspensive/react'
import * as Sentry from '@sentry/react'

function App() {
  return (
    <ErrorBoundary.Observer onError={Sentry.captureReactException}>
      <ErrorBoundary
        fallback={<FallbackA />}
        onError={(error) => console.log('also called!', error)} // not overridden by Observer
      >
        <YourComponent />
      </ErrorBoundary>
    </ErrorBoundary.Observer>
  )
}
// When an error occurs:
// 1. ErrorBoundary's onError: console.log('also called!', error)
// 2. ErrorBoundary.Observer's onError: Sentry.captureReactException(error, info)
```

### Features

- **Bubble order**: Execution order is local `onError` → inner Observer → outer Observer
- **Nestable**: Multiple `ErrorBoundary.Observer` components can be nested, and all will be called
- **`@experimental`**: Marked as an experimental feature

### Changes

- `packages/react/src/ErrorBoundary.tsx`: Add `ErrorBoundary.Observer` static member and `ErrorObserverContext`
- `packages/react/src/ErrorBoundary.spec.tsx`: Add 4 test cases
- `docs/suspensive.org/src/content/{en,ko}/docs/react/ErrorBoundary.mdx`: Add documentation (comparison table, Observer section, Sentry migration guide)


.changeset/early-hornets-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+"@suspensive/react": minor
+---
+
+feat(ErrorBoundary): add ErrorBoundary.Observer
```
manudeli:
```suggestion
feat(react): add ErrorBoundary.Observer
```

kangju2000 (commented):
**1. The name "Observer" might not align with the actual behavior**

"Observer" tends to evoke the observer pattern (subscribe/unsubscribe), but the actual implementation is a Context Provider — it propagates an error callback through Context for ErrorBoundary to consume. There might be a gap between the mental model the name suggests and how it actually works.

**2. Could a wrapper component achieve a similar result?**

If the core motivation is "apply the same onError to all ErrorBoundaries," a wrapper seems like it could solve this just as well:

```tsx
const AppErrorBoundary = (props) => (
  <ErrorBoundary onError={Sentry.captureReactException} {...props} />
)
```

Observer adds `useContext` + `useCallback` to every ErrorBoundary instance for this purpose, and this cost is incurred even when Observer isn't used — I'm curious whether this tradeoff is considered acceptable.

**3. If the goal is default props, would DefaultPropsProvider be a more natural fit?**

Since suspensive already has `DefaultPropsProvider`, adding a default `onError` there might feel more consistent with the existing API. Extending a pattern users already know could be preferable to introducing a new component — what do you think?

gwansikk (approved):
LGTM

PR #1918, opened 2026-03-17
chore: remove banner option from tsdown configurations and add 'use client' directive to multiple React components

manudeli:


packages/react-query/tsdown.config.ts
```diff
@@ -1,10 +1,4 @@
...
 import { defineConfig } from 'tsdown'
 
-export default defineConfig([
-  {
-    ...options,
-    banner: { js: undefined },
-  },
-  scriptOptions,
-])
+export default defineConfig([options, scriptOptions])
```
manudeli, line 4: consistent config

packages/react-query-5/tsdown.config.ts
```diff
@@ -1,7 +1,4 @@
 import { options } from '@suspensive/tsdown'
 import { defineConfig } from 'tsdown'
 
-export default defineConfig({
-  ...options,
-  banner: undefined,
-})
+export default defineConfig(options)
```
manudeli, line 4: consistent config

packages/react-query-4/tsdown.config.ts
```diff
@@ -1,7 +1,4 @@
 import { options } from '@suspensive/tsdown'
 import { defineConfig } from 'tsdown'
 
-export default defineConfig({
-  ...options,
-  banner: undefined,
-})
+export default defineConfig(options)
```
manudeli, line 4: consistent config

PR #1922, opened 2026-03-17
refactor(jotai): rename AtomValue to ReadAtom

zaewc:
# Overview

  `AtomValue` is the only component in `@suspensive/jotai` that doesn't follow the consistent naming pattern.

  - `Atom` → `[Target]`
  - `SetAtom` → `[Verb][Target]` 
  - `AtomValue` → `[Target][Attribute]`

  This PR renames `AtomValue` to `ReadAtom`, aligning with the `[Verb][Target]` pattern used by `SetAtom`.

  ## Changes

  - `@suspensive/jotai`: `AtomValue` -> `ReadAtom`
  - Updated tests, docs, and related references

  ## PR Checklist

### DIFF
--- a/.changeset/rename-read-atom.md
+++ b/.changeset/rename-read-atom.md
@@ -1,0 +1,5 @@
+---
+"@suspensive/jotai": major
+---
+
+Rename `AtomValue` to `ReadAtom` for naming consistency


976520:
I'd like to raise a question about whether ReadAtom is the right name here.                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                             
  The `[Verb][Target]` pattern does complete a clean symmetry — `Atom` / `ReadAtom` / `SetAtom` maps 1:1 to `useAtom` / `useReadAtom` / `useSetAtom`. But Jotai's official API uses `useAtomValue`, so `AtomValue` was actually closer to upstream Jotai's naming convention. 
  
  Since `useReadAtom` is a hook provided by `@suspensive/jotai` itself, the consistency argument is really about aligning with suspensive's own internal layer rather than Jotai itself. 
  
  Given this is a breaking change, wanted to make sure this tradeoff is intentional before merging 🥰

PR #1935, opened 2026-03-27
fix(react-query): guide users to install @suspensive/react-query-4 or @suspensive/react-query-5 directly via npm aliases

manudeli:


docs/suspensive.org/src/content/en/docs/react-query/_meta.tsx
```diff
@@ -3,9 +3,6 @@ import type { MetaRecord } from 'nextra'
 export default {
   motivation: { title: 'Motivation' },
   installation: { title: 'Installation' },
-  'tanstack-query-compatibility': {
-    title: 'TanStack Query v4 & v5 Support',
-  },
```
manudeli, line 8: Remove this page

docs/suspensive.org/src/content/en/docs/react-query/installation.mdx
```diff
@@ -51,27 +51,30 @@ Suspensive Libraries are optimized for modern browsers. It is compatible with th
...
-@suspensive/react-query is available on npm.
-From version 2.2.0 onwards, it supports both v4 and v5 of @tanstack/react-query. Depending on the version of @tanstack/react-query specified in your package.json dependencies, the appropriate version of @suspensive/react-query will be used automatically.
-
-To use the latest stable version, run the following command.
+<Tabs items={['@tanstack/react-query v5', '@tanstack/react-query v4']}>
+<Tabs.Tab>
 
 ```shell npm2yarn
-npm install @suspensive/react-query @suspensive/react @tanstack/react-query
+npm install @suspensive/react-query@npm:@suspensive/react-query-5 @tanstack/react-query@5
```
manudeli, line 60: Update this installation script

docs/suspensive.org/src/content/en/docs/react-query/installation.mdx
```diff
@@ -51,27 +51,30 @@ Suspensive Libraries are optimized for modern browsers. It is compatible with th
...
+npm install @suspensive/react-query@npm:@suspensive/react-query-4 @tanstack/react-query@4
 ```
 
+</Tabs.Tab>
+</Tabs>
+
 <Callout type='info'>
 
-@suspensive/react-query internally uses @tanstack/react-query v4 or v5 as peerDependencies. Therefore, ensure that the correct version of @tanstack/react-query is installed in your package.json dependencies. The reasons are explained in [Why Use This?](/docs/react-query/motivation).
+By using npm aliases, the import path remains `@suspensive/react-query` across your codebase. When upgrading your @tanstack/react-query version, you only need to change the alias in package.json — no code changes required.
```
manudeli, line 77: Cause why we recommend using npm aliases

packages/react-query/src/bin/postinstall.ts
```diff
@@ -14,6 +15,27 @@ switch (major) {
...
+  {
+    "dependencies": {
+      "@suspensive/react-query": "npm:@suspensive/react-query-${major}@^${suspensivePackageJson.version}"
+    }
+  }
+
+  When upgrading @tanstack/react-query, just change the alias in package.json — no code changes required.
+
+  See: https://suspensive.org/docs/react-query/installation
+`)
```
manudeli: warning! when we postinstall

manudeli (commented):
@gwansikk Check this Pull Request please

packages/react-query/src/bin/postinstall.ts
gwansikk: 👍👍

docs/suspensive.org/src/content/en/docs/react-query/_meta.tsx
gwansikk: Suggestion: Rather than removing it immediately, how about updating the content to note that this approach is deprecated and officially no longer supported, then keeping it around for a while?

manudeli: In my opinion, since we won't be maintaining this content going forward, removing it helps highlight the recommended installation path more clearly. Also, the CLI and auto-detection docs would contradict the new npm alias approach, which could confuse users. The postinstall warning already covers the migration guidance for existing users

manudeli: @gwansikk It honestly hurts to remove something you worked so hard on — the CLI and compatibility docs were really well done. But with the direction shifting to npm aliases, keeping them around would send mixed signals to users

gwansikk: I Agreed! No problem at all. It was just a simple suggestion on my part.

packages/react-query/src/bin/postinstall.ts
gwansikk: Great idea! How about also surfacing it in the CLI? I think this message should be shown to users as frequently as possible

manudeli: I reflected your review in [09e5019](https://github.com/toss/suspensive/pull/1935/commits/09e50190ef231fec8f316456a7b7e64cb97d5d0e)

Issue #1895, opened 2026-01-30
[Feature]: QueriesHydration disable retry all queries


manudeli:
Idea by @bluejoyq 

### Package Scope

@suspensive/react-query-4, @suspensive/react-query-5

### Description

QueriesHydration disable retry all queries

manudeli:
@bluejoyq Can I assign you to resolve this issue?

PR #1948, opened 2026-04-07
fix(react-query-4,react-query-5): disable retry in QueriesHydration s…

rjsdnql123:
Hi Suspensive team,
Thanks for the great library!

# Overview
Issue #1895

### What

Force retry: 0 when calling fetchQuery and fetchInfiniteQuery in QueriesHydration.

### Why

If retries occur during server-side prefetching, the server component rendering becomes blocked for each retry attempt.

### How
Explicitly set `retry` to `0` for all query calls:

fetchQuery({ ...query, retry: 0 })
fetchInfiniteQuery({ ...query, retry: 0 })

## PR Checklist

### DIFF
--- a/packages/react-query-4/src/QueriesHydration.tsx
+++ b/packages/react-query-4/src/QueriesHydration.tsx
@@ -117,7 +117,9 @@ export async function QueriesHydration({
   try {
     const queriesPromise = Promise.all(
       queries.map((query) =>
-        'getNextPageParam' in query ? queryClient.fetchInfiniteQuery(query) : queryClient.fetchQuery(query)
+        'getNextPageParam' in query
+          ? queryClient.fetchInfiniteQuery({ ...query, retry: 0 })
+          : queryClient.fetchQuery({ ...query, retry: 0 })
       )
     )
     await (timeoutController != null ? Promise.race([queriesPromise, timeoutController.promise]) : queriesPromise)
--- a/packages/react-query-5/src/QueriesHydration.tsx
+++ b/packages/react-query-5/src/QueriesHydration.tsx
@@ -113,7 +113,9 @@ export async function QueriesHydration({
   try {
     const queriesPromise = Promise.all(
       queries.map((query) =>
-        'getNextPageParam' in query ? queryClient.fetchInfiniteQuery(query) : queryClient.fetchQuery(query)
+        'getNextPageParam' in query
+          ? queryClient.fetchInfiniteQuery({ ...query, retry: 0 })
+          : queryClient.fetchQuery({ ...query, retry: 0 })
       )
     )
     await (timeoutController != null ? Promise.race([queriesPromise, timeoutController.promise]) : queriesPromise)


packages/react-query-4/src/QueriesHydration.tsx
```diff
@@ -117,7 +117,9 @@ export async function QueriesHydration({
   try {
     const queriesPromise = Promise.all(
       queries.map((query) =>
-        'getNextPageParam' in query ? queryClient.fetchInfiniteQuery(query) : queryClient.fetchQuery(query)
+        'getNextPageParam' in query
+          ? queryClient.fetchInfiniteQuery({ ...query, retry: 0 })
+          : queryClient.fetchQuery({ ...query, retry: 0 })
```
marshallku, line 122: How about adding this option just on the server side?
Current solution would work both client and server side.

rjsdnql123: `QueriesHydration` is an async function component that runs only on the server. If it is used in a client component (`'use client'`), it will cause an async component rendering error.

Therefore, `retry: 0` is applied only on the server side, while `useQuery` on the client follows its own retry configuration:

1. Server: `fetchQuery({ retry: 0 })` → stops immediately after the first failure  
2. Client: `skipSsrOnError` → `useQuery` mounts and retries based on its own retry settings  

I might have misunderstood the review, so please feel free to correct me if anything is off.
I also tested this scenario, and confirmed that retries on the client are working as expected. If you could share the specific case you're referring to, I’d be happy to test it as well.

rjsdnql123: @marshallku : )

marshallku: Sorry for being late @rjsdnql123 
If so, I think this is quite proper change.
How about you @manudeli

gwansikk:
I think if a user explicitly sets retry, we should respect that setting. TanStack Query takes the same stance — it defaults retry to 0 on the server, but if the user provides an explicit value, it follows that instead.

That said, QueriesHydration sits one layer of abstraction above TanStack Query — it always runs on the server (since it's an async component) and acts as a wrapper around fetchQuery. Given this narrower context, I do think there's room to justify enforcing SSR safety more proactively here.

Still, silently overriding a user's explicit retry value can make debugging harder. What if we exposed this as a prop instead, so the user can opt in consciously? For example, a `disableRetryOnServer` prop would let users make an intentional choice about the behavior they want.

What do you think? @manudeli @kangju2000 @marshallku 

<details>
<summary>KOR</summary>
제 생각에는 사용자가 `retry`를 명시했다면 이를 이행하는 것이 맞다고 생각합니다.
TanStack Query도 동일한 설정이에요. 서버에서 retry default를 0으로 보장하지만, 사용자가 명시한 경우엔 그 설정을 따릅니다.

다만 `QueriesHydration`은 TanStack Query 위에 한 단계 추상화되어 있고, 항상 서버에서만 실행되며 (async component) `fetchQuery`를 호출하는 wrapper라는 컨텍스트를 가지고 있어요. 그래서 SSR 안전성을 더 적극적으로 보장하는 게 정당화될 여지는 있다고 생각해요.

그럼에도 사용자 명시값을 override하는 건 디버깅을 어렵게 만들 수 있어서, `props`로 사용자가 선택할 수 있게 노출하는 방향은 어떨까요? 예를 들어 `disableRetryOnServer` 같은 `prop`을 두면 사용자가 의식하고 의도한 대로 동작을 선택할 수 있습니다. 

어떻게 생각하세요?
</details>

PR #1950, opened 2026-04-13
test(jotai): add options tests for Atom, AtomValue, and SetAtom

JeongHwan-dev:
# Overview

## Summary                                                                                                                                             
                                                                                                                                                         
Add tests for the `options` prop across all three components (`Atom`, `AtomValue`, `SetAtom`) in `@suspensive/jotai`.                                  
                                                                                                                                                       
All three components accept an `options` prop that is passed directly to the underlying jotai hooks (`useAtom`, `useAtomValue`, `useSetAtom`), but there were no tests verifying this behavior. Since the `atom` prop is already tested with the same pass-through pattern, the `options` prop should be tested as well for consistency.                                                                                                                        
                                                                                                                                                         
## Changes                                                                                                                                           

Two test cases added per component:                                                                                                                    

**1. Store pass-through** — verifies that the component reads/writes from the specified store                                                          
                                                                                                                                                   
**2. Store isolation** — verifies that writing to `storeA` does not affect `storeB` when using the same atom                                           
                                                                                                                                                       
| File | Added Tests |                                                                                                                                 
|------|------------|                                                                                                                                
| `Atom.spec.tsx` | `should read and write value from the specified store via options` |                                                             
| | `should isolate state between different stores via options` |                                                                                      
| `AtomValue.spec.tsx` | `should read value from the specified store via options` |                                                                    
| | `should isolate state between different stores via options` |                                                                                      
| `SetAtom.spec.tsx` | `should write to the specified store via options` |                                                                             
| | `should isolate state between different stores via options` | 

<!--
    A clear and concise description of what this pr is about.
 -->

## PR Checklist

### DIFF
--- a/packages/jotai/src/Atom.spec.tsx
+++ b/packages/jotai/src/Atom.spec.tsx
@@ -1,5 +1,5 @@
 import { act, fireEvent, render, screen } from '@testing-library/react'
-import { atom } from 'jotai'
+import { atom, createStore } from 'jotai'
 import { Suspense } from 'react'
 import { Atom } from './Atom'
 import { sleep } from './test-utils'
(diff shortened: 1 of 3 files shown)


manudeli (approved):
Thanks!

PR #1953, opened 2026-04-14
fix(suspensive.org/Sandpack): exclude custom props from 'SandpackPreview' spread to prevent DOM warnings

sukvvon:
# Overview

- Destructure custom props (`layout`, `showConsole`, `showConsoleButton`) from `previewOptions` before spreading to `SandpackPreview`
- These props are not part of `PreviewProps` but were passed to the DOM via `{...props.previewOptions}`, causing React warnings:
  - `showConsoleButton` prop on a DOM element
  - `showConsole` prop on a DOM element
  
### AS-IS

image
image
image

### TO-BE

image

## PR Checklist


kangju2000 (approved):
LGTM

PR #1958, opened 2026-04-30
fix: Inconsistent UI Rendering When System is in Light Mode

### LINKED ISSUE #1957: [Bug]: Inconsistent UI Rendering When System is in Light Mode
### Package Scope

etc

### Bug description

After Light Mode was removed and the default background became dark, a color mismatch occurs when the system is set to Light Mode.

Some UI elements (text, table, or highlight colors) still appear to follow Light Mode styles, which results in poor contrast and reduced readability against the dark background. (See screenshot)

I’d be happy to work on this if it’s considered an issue.

| Environment | Screenshot |
|------------|-----------|
| Light Mode | Image|
| Dark Mode |  Image|

### Expected behavior

The UI should remain visually consistent regardless of the system theme.
If the application is now fully Dark Mode-based, all components should follow the same color scheme and not inherit Light Mode styles.

### To Reproduce

_No response_

### Possible Solution

_No response_

### etc.

_No response_

myeong-jae-hwi:
# Overview
After removing Light Mode and setting the default background to dark, a color mismatch occurs when the system is set to Light Mode.

Some UI elements (such as text, tables, and highlights) still follow Light Mode styles, resulting in poor contrast and reduced readability against the dark background.

Tailwind’s dark: variant applies styles based on the OS color scheme. As a result, dark: classes were ignored when the system was in Light Mode.

Since next-themes already applies a dark class to the <html> element via forcedTheme, we introduced a custom dark variant so that Tailwind references this class instead. This ensures that dark mode styles are always applied regardless of the system setting.

## PR Checklist

### DIFF
--- a/docs/suspensive.org/src/app/[lang]/styles.css
+++ b/docs/suspensive.org/src/app/[lang]/styles.css
@@ -1,6 +1,9 @@
 @import 'tailwindcss';
 @import 'nextra-theme-docs/style.css';
 
+/* tailWind가 OS의 테마 설정이 아닌, HTML 내의 dark 클래스를 참조하도록 설정 */
+@custom-variant dark (&:where(.dark, .dark *));
+
 .scrollbar-none {
   -ms-overflow-style: none;
   scrollbar-width: none;


manudeli (approved):

manudeli:
Thanks! 🙏

